PR #7, opened 2022-06-27
Swapchain example fixes

deprilula28:
Sister PR: https://github.com/Devsh-Graphics-Programming/Nabla/pull/368

Examples to retest now that clip control is no more

- [x] 02
- [ ] 03 (doesn't run)
- [ ] 05 (doesn't run)
- [x] 06
- [x] 07
- [x] 09
- [x] 10
- [x] 14
- [ ] 20 (left unfixed)
- [x] 21
- [x] 23
- [x] 27
- [x] 29 (a single lonely dot on opengl)
- [x] 33
- [x] 34
- [x] 35
- [x] 42
- [x] 43
- [x] 46
- [x] 47
- [ ] 49 (doesn't exist?)
- [x] 52 
- [x] 53 (a single lonely dot on opengl)
- [ ] 54 (doesn't run)
- [x] 55
- [x] 56 (my gpu doesnt support rt) @Erfan-Ahmadi


42.FragmentShaderPathTracer/common.glsl
```diff
@@ -678,10 +680,12 @@ bool closestHitProgram(in uint depth, in uint _sample, inout Ray_t ray, inout nb
 
 void main()
 {
+    const ivec2 imageExtents = nbl_glsl_swapchain_transform_preTransformExtents(cameraData.swapchainTransform, imageSize(outImage));
     const ivec2 coords = getCoordinates();
-    const vec2 texCoord = getTexCoords();
+    vec2 texCoord = (vec2(nbl_glsl_swapchain_transform_preTransform(cameraData.swapchainTransform, coords, imageSize(outImage))) / vec2(imageExtents));
+    texCoord.y = 1.0 - texCoord.y;
```
Erfan-Ahmadi, line 687: Why do we need to flip y again if we did everything correctly?

07.SubpassBaking/main.cpp
```diff
@@ -522,7 +530,10 @@ class SubpassBaking : public ApplicationBase
         // update camera
         {
             const auto& viewMatrix = camera.getViewMatrix();
-            const auto& viewProjectionMatrix = camera.getConcatenatedMatrix();
+            const auto& viewProjectionMatrix = matrix4SIMD::concatenateBFollowedByAPrecisely(
```
Erfan-Ahmadi, line 533: Why don't we do this for example 6 also?

42.FragmentShaderPathTracer/common.glsl
devshgraphicsprogramming: only one transform should exist

In an example that renders vertices to screen you apply it to the projection matrix (given triangle vertices you map them to screen). Remember rasterization is a scatter op.

However in an example such as 42, where you're casting rays, you're finding out worldspace positions for screenspace pixels so transform needs to be applied in reverse. Raycasting/post process is a gather op.

Erfan-Ahmadi: @deprilula28 can you comment what @devshgraphicsprogramming wrote
Even in a comment above this line will do

42.FragmentShaderPathTracer/common.glsl
```diff
@@ -686,10 +688,14 @@ bool closestHitProgram(in uint depth, in uint _sample, inout Ray_t ray, inout nb
...
+    const ivec2 imageExtents = nbl_glsl_swapchain_transform_preTransformExtents(cameraData.swapchainTransform, imageSize(outImage));
     const ivec2 coords = getCoordinates();
-    const vec2 texCoord = getTexCoords();
+    // In an example that renders vertices to screen you apply transform it to the projection matrix (given triangle vertices you map them to screen). Remember rasterization is a scatter op.
+    // However in an example such as this, where you're casting rays, you're finding out worldspace positions for screenspace pixels so transform needs to be applied in reverse. Raycasting/post process is a gather op.
+    vec2 texCoord = (vec2(nbl_glsl_swapchain_transform_preTransform(cameraData.swapchainTransform, coords, imageSize(outImage))) / vec2(imageExtents));
+    texCoord.y = 1.0 - texCoord.y;
 
-    if (false == (all(lessThanEqual(ivec2(0),coords)) && all(greaterThan(imageSize(outImage),coords)))) {
+    if (false == (all(lessThanEqual(ivec2(0),coords)) && all(greaterThan(imageExtents,coords)))) {
```
Erfan-Ahmadi, line 698: Should we use `texCoord` here instead of `coords`?
Assuming we have a
```
transform = NBL_GLSL_SWAPCHAIN_TRANSFORM_E_ROTATE_90
[width,height] = [10, 20]
coords range are from [0, 0] to [10, 20] (right?)
```
then 
`nbl_glsl_swapchain_transform_preTransformExtents` will give us `[20, 10]` and
then we will only render to [10, 10] 

Of course it doesn't change anything right now when we only test HORIZONTAL_180_MIRROR

27.PLYSTLDemo/main.cpp
```diff
@@ -149,7 +149,7 @@ APP_CONSTRUCTOR(PLYSTLDemo)
 		const auto swapchainImageUsage = static_cast<asset::IImage::E_USAGE_FLAGS>(asset::IImage::EUF_COLOR_ATTACHMENT_BIT);
 		const video::ISurface::SFormat surfaceFormat(asset::EF_R8G8B8A8_SRGB, asset::ECP_COUNT, asset::EOTF_UNKNOWN);
 
-		CommonAPI::InitWithDefaultExt(initOutput, video::EAT_VULKAN/*Vulkan doesn't work yet*/, "plystldemo", WIN_W, WIN_H, SC_IMG_COUNT, swapchainImageUsage, surfaceFormat, nbl::asset::EF_D32_SFLOAT);
+		CommonAPI::InitWithDefaultExt(initOutput, video::EAT_OPENGL/*Vulkan doesn't work yet*/, "plystldemo", FRAMES_IN_FLIGHT, WIN_W, WIN_H, SC_IMG_COUNT, swapchainImageUsage, surfaceFormat, nbl::asset::EF_D32_SFLOAT);
```
devshgraphicsprogramming, line 152: make Vulkan work ;)

42.FragmentShaderPathTracer/main.cpp
```diff
@@ -387,6 +387,7 @@ int main()
 	struct SBasicViewParametersAligned
 	{
 		SBasicViewParameters uboData;
+		uint32_t surfaceTransform;
```
Erfan-Ahmadi, line 390: Remove this, we don't use this in the shader anymore

PR #8, opened 2022-07-11
Resizing Maintenance

deprilula28:
Maintenance for resizing, TODOS:

Bigger maintenance crap:
- [ ]  redo surfaceFormat into a list (in order of preference) of "acceptable" formats (format, colorspace, OETF) https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/blob/example-fixes/common/CommonAPI.h#L944
- [ ]  fix this https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/blob/example-fixes/common/CommonAPI.h#L207
Small maintenance crap:
- [ ]  the sc argument is unused https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/blob/example-fixes/common/CommonAPI.h#L1712
- [ ]  add a comment here https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/blob/example-fixes/common/CommonAPI.h#L1553 that its just until OpenGL backend reports swapchain formats properly 
- [ ]  whatever is not templated, should move to a CommonAPI.cpp file (before everything was templated)
- [ ]  Change the signature of this function https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/blob/example-fixes/common/CommonAPI.h#L933 to be static InitOutput Init(InitParams&& params);
- [ ]  turn all existing arguments into InitParams struct members with nice default values (makes any further refactors easier across 50 examples)
    - [ ]  some members from InitOutput would need to be added to InitParams like window, windowCb and windowManager (android does stuff weird)
    - [ ]  add a new, last smart_refctd_ptr<system::ILogger> m_logger=nullptr member and solve this comment https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/blob/example-fixes/common/CommonAPI.h#L956-958
- [ ]  use your format promotion to promote depth format in CommonAPI https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/blob/example-fixes/common/CommonAPI.h#L1581 and https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/blob/example-fixes/common/CommonAPI.h#L1641
- [ ]  type returned should be a core::smart_refctd_dynamic_array of swapchain image count instead of fixed size std::array in https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/blob/example-fixes/common/CommonAPI.h#L1638
- [ ]  remove the default arg here https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/blob/example-fixes/common/CommonAPI.h#L1762
- [ ]  use core::bitflag here https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/blob/example-fixes/common/CommonAPI.h#L1787
- [ ]  log failures nicely (example will currently blow up if any thing that can return nullptr or error does so)

### DIFF
--- a/common/CommonAPI.h
+++ b/common/CommonAPI.h
@@ -806,6 +806,9 @@ public:
 		FeatureType* features = nullptr;
 	};
 
+	struct InitArgs {
+	};
+
 	struct InitOutput
 	{
 		enum E_QUEUE_TYPE


common/CommonAPI.h
```diff
@@ -734,23 +734,68 @@ class CommonAPI
...
 		SFeatureRequest<nbl::video::IAPIConnection::E_FEATURE> requiredInstanceFeatures = {};
 		SFeatureRequest<nbl::video::IAPIConnection::E_FEATURE> optionalInstanceFeatures = {};
 		SFeatureRequest<nbl::video::ILogicalDevice::E_FEATURE> requiredDeviceFeatures = {};
 		SFeatureRequest<nbl::video::ILogicalDevice::E_FEATURE> optionalDeviceFeatures = {};
 		
-		nbl::asset::IImage::E_USAGE_FLAGS swapchainImageUsage = nbl::asset::IImage::E_USAGE_FLAGS::EUF_NONE;
-		nbl::core::vector<nbl::video::ISurface::SFormat> acceptableSurfaceFormats = {
-			nbl::video::ISurface::SFormat(nbl::asset::EF_R8G8B8A8_SRGB, nbl::asset::ECP_SRGB, nbl::asset::EOTF_sRGB),
-			nbl::video::ISurface::SFormat(nbl::asset::EF_B8G8R8A8_SRGB, nbl::asset::ECP_SRGB, nbl::asset::EOTF_sRGB)
+		nbl::asset::IImage::E_USAGE_FLAGS swapchainImageUsage = nbl::asset::IImage::E_USAGE_FLAGS::EUF_COLOR_ATTACHMENT_BIT;
```
devshgraphicsprogramming, line 743: don use complex containters such as vectors, use pointer + count, two pointers and begin/end iterators or an invalid value terminated list

store the default values as `constexpr static inline std::array` or plain array

common/CommonAPI.h
```diff
@@ -734,23 +734,68 @@ class CommonAPI
...
+		};
+
 		nbl::asset::E_FORMAT depthFormat = nbl::asset::EF_UNKNOWN;
 
 		nbl::core::smart_refctd_ptr<nbl::ui::IWindow> window = nullptr;
 		nbl::core::smart_refctd_ptr<nbl::ui::IWindowManager> windowManager = nullptr;
 		nbl::core::smart_refctd_ptr<CommonAPIEventCallback> windowCb = nullptr;
 		nbl::core::smart_refctd_ptr<nbl::system::ILogger> logger = nullptr;
+
+		bool headlessCompute = false;
```
devshgraphicsprogramming, line 766: no need for this bool, you should be able to deduce from requiring the swapchain & surface features

common/CommonAPI.h
```diff
@@ -1566,10 +1485,17 @@ class CommonAPI
...
+	) -> std::array<nbl::core::smart_refctd_ptr<nbl::video::IGPUFramebuffer>, InitOutput::MaxSwapChainImageCount>
 	{
 		using namespace nbl;
-		bool useDepth = asset::isDepthOrStencilFormat(depthFormat);
+
+		auto depthFormat = device->getPhysicalDevice()->promoteImageFormat(
+			{ baseDepthFormat, nbl::video::IPhysicalDevice::SFormatImageUsage(nbl::asset::IImage::EUF_DEPTH_STENCIL_ATTACHMENT_BIT) },
+			nbl::asset::IImage::ET_OPTIMAL
+		);
+		bool useDepth = depthFormat != nbl::asset::EF_UNKNOWN;
```
devshgraphicsprogramming, line 1499: figure this out before promotion, we don't want to drop the usage of a depth buffer just because someone couldn't promote

02.ComputeShader/main.cpp
```diff
@@ -102,35 +102,37 @@ class ComputeShaderSampleApp : public ApplicationBase
...
-		CommonAPI::InitOutput initOutput;
-		initOutput.window = core::smart_refctd_ptr(window);
-		CommonAPI::InitWithDefaultExt(
-			initOutput,
-			video::EAT_VULKAN,
-			"02.ComputeShader",
-			FRAMES_IN_FLIGHT, WIN_W, WIN_H, 3u,
-			swapchainImageUsage,
-			surfaceFormat);
+		std::vector<asset::E_FORMAT> acceptableSurfaceFormats = { asset::EF_B8G8R8A8_UNORM };
```
devshgraphicsprogramming, line 105: `const std:array` would be nicer

02.ComputeShader/main.cpp
```diff
@@ -102,35 +102,37 @@ class ComputeShaderSampleApp : public ApplicationBase
...
+		CommonAPI::InitParams initParams;
+		initParams.window = core::smart_refctd_ptr(window);
+		initParams.apiType = video::EAT_VULKAN;
+		initParams.appName = { "02.ComputeShader" };
+		initParams.framesInFlight = FRAMES_IN_FLIGHT;
+		initParams.windowWidth = WIN_W;
+		initParams.windowHeight = WIN_H;
+		initParams.scImageCount = 3u;
+		initParams.swapchainImageUsage = swapchainImageUsage;
+		initParams.acceptableSurfaceFormats = &acceptableSurfaceFormats[0];
```
devshgraphicsprogramming, line 116: `.data()`

02.ComputeShader/main.cpp
```diff
@@ -113,7 +113,7 @@ class ComputeShaderSampleApp : public ApplicationBase
 		initParams.windowHeight = WIN_H;
 		initParams.scImageCount = 3u;
 		initParams.swapchainImageUsage = swapchainImageUsage;
-		initParams.acceptableSurfaceFormats = &acceptableSurfaceFormats[0];
+		initParams.acceptableSurfaceFormats = &acceptableSurfaceFormats.data()[0];
```
devshgraphicsprogramming, line 116: just `acceptableSurfaceFormats.data()`

common/CommonAPI.h
```diff
@@ -867,176 +474,111 @@ class CommonAPI
...
+			params.windowCb->setInputSystem(nbl::core::smart_refctd_ptr(result.inputSystem));
 #else
-		result.windowCb = nbl::core::smart_refctd_ptr<EventCallback>((CommonAPIEventCallback*)result.window->getEventCallback());
-		result.windowCb->setInputSystem(nbl::core::smart_refctd_ptr(result.inputSystem));
+			params.windowCb = nbl::core::smart_refctd_ptr<EventCallback>((CommonAPIEventCallback*)params.window->getEventCallback());
+			params.windowCb->setInputSystem(nbl::core::smart_refctd_ptr(result.inputSystem));
 #endif
 		}
 
 		if constexpr (gpuInit)
```
devshgraphicsprogramming, line 529: this surface stuff and API connection you can probably move to cpp

common/CommonAPI.h
```diff
@@ -806,6 +369,53 @@ class CommonAPI
...
+
+		constexpr bool isHeadlessCompute()
+		{
+			return swapchainImageUsage == nbl::asset::IImage::E_USAGE_FLAGS(0);
+		}
+	};
+
 	struct InitOutput
 	{
 		enum E_QUEUE_TYPE
```
Erfan-Ahmadi, line 421: Let's move these outside, CommonAPI::EQT_GRAPHICS looks much better than CommonAPI::InitOutput::EQT_GRAPHICS

devshgraphicsprogramming: its not about looks, its about encapsulation, does anyone else exceptthings that deal with `InitOutput` need to deal with this enum, if not, it should stay in the deepest scope possible

common/CommonAPI.cpp
```diff
@@ -1,5 +1,6 @@
 
 #include "CommonAPI.h"
+#include "nbl/video/CVulkanSwapchain.h"
```
devshgraphicsprogramming, line 3: make `nbl/video/video.h` include it

02.ComputeShader/main.cpp
```diff
@@ -99,37 +118,123 @@ class ComputeShaderSampleApp : public ApplicationBase
...
+		std::array<asset::E_FORMAT, 1> acceptableSurfaceFormats = { asset::EF_B8G8R8A8_UNORM };
+
+		CommonAPI::InitParams initParams;
+		initParams.window = core::smart_refctd_ptr(window);
+		initParams.apiType = video::EAT_VULKAN;
+		initParams.appName = { "02.ComputeShader" };
+		initParams.framesInFlight = FRAMES_IN_FLIGHT;
+		initParams.windowWidth = windowWidth;
+		initParams.windowHeight = windowHeight;
+		initParams.scImageCount = 3u;
```
Erfan-Ahmadi, line 214: please rename this to `swapchainImageCount`

common/CommonAPI.cpp
```diff
@@ -0,0 +1,1030 @@
...
+		assert(fbo->begin()[i]);
+	}
+	return fbo;
+}
+
+bool CommonAPI::createSwapchain(
+	const nbl::core::smart_refctd_ptr<nbl::video::ILogicalDevice>&& device,
+	nbl::video::ISwapchain::SCreationParams& params,
+	uint32_t width, uint32_t height,
+	// nullptr for initial creation, old swapchain for eventual resizes
```
Erfan-Ahmadi, line 663: this comment makes no sense unless we returned `nbl::core::smart_refctd_ptr<nbl::video::ISwapchain>` instead of `bool`

common/CommonAPI.cpp
```diff
@@ -0,0 +1,1030 @@
...
+	nbl::core::smart_refctd_ptr<nbl::video::ISwapchain>& swapchain
+)
+{
+	auto oldSwapchain = swapchain;
+
+	nbl::video::ISwapchain::SCreationParams paramsCp = params;
+	paramsCp.width = width;
+	paramsCp.height = height;
+	paramsCp.oldSwapchain = oldSwapchain;
```
Erfan-Ahmadi, line 672: 1. Why would the user provide the params when he can just set these in the params?
seems redundant.

The only use for this function I see is hiding the API things.
2. I also don't understand why we removed that from ILogicalDevice and now don't have an API agnostic `createSwapchain` function in Nabla, Can you clarify?

02.ComputeShader/main.cpp
```diff
@@ -5,13 +5,18 @@
 
 using namespace nbl;
 
-class ComputeShaderSampleApp : public ApplicationBase
+// TODO better way of handling this
```
Erfan-Ahmadi, line 8: Agreed this shit looks usgly, specially since it's the first thing you see in the example and Examples should inherit from hit.

02.ComputeShader/main.cpp
```diff
@@ -99,37 +135,96 @@ class ComputeShaderSampleApp : public ApplicationBase
...
+		initParams.appName = { "02.ComputeShader" };
+		initParams.framesInFlight = FRAMES_IN_FLIGHT;
+		initParams.windowWidth = windowWidth;
+		initParams.windowHeight = windowHeight;
+		initParams.swapchainImageCount = 3u;
+		initParams.swapchainImageUsage = swapchainImageUsage;
+		initParams.acceptableSurfaceFormats = acceptableSurfaceFormats.data();
+		initParams.acceptableSurfaceFormatCount = acceptableSurfaceFormats.size();
+		auto initOutput = CommonAPI::InitWithDefaultExt<true, ResizeEventHandler>(std::move(initParams));
+		((ResizeEventHandler*) initParams.windowCb.get())->callback = this;
```
Erfan-Ahmadi, line 208: dynamic_cast

common/CommonAPI.cpp
deprilula28: This function takes a reference to the swapchain, places the current value into oldSwapchaib 1and writes the new value into the reference. The first time it's ran it'll be null, then next times it will have the old swapchain

common/CommonAPI.cpp
deprilula28: 1. That's more of a convenience thing, the examples don't really touch the parameters outside of moving them around

2. That was done for decoupling swapchain from logical device, do you want there to be a function that creates it?

Erfan-Ahmadi: > That was done for decoupling swapchain from logical device
Ok but can you remind me why we want them decoupled?

common/CommonAPI.cpp
Erfan-Ahmadi: Ok i get it now it's a "Reference" to `smart_refctd_ptr" and it is null on the first run.

common/CommonAPI.h
deprilula28: kind of, this gets used every time an example wants to fetch a queue

Erfan-Ahmadi: Yea places that happen much later than CommonAPI::Init has happened. you can see things like queues[CommonAPI::InitOutput::EQT_GRAPHICS]

02.ComputeShader/main.cpp
```diff
@@ -43,11 +50,40 @@ class ComputeShaderSampleApp : public ApplicationBase
...
+
+	struct CSwapchainResources : public CommonAPI::IRetiredSwapchainResources
+	{
+		nbl::core::smart_refctd_ptr<nbl::video::IGPUImageView> oldImageView = nullptr;
+		nbl::core::smart_refctd_ptr<nbl::video::IGPUImage> oldImage = nullptr;
+
+		~CSwapchainResources() override {}
+	};
+
+	struct ResizeEventHandler : public CommonAPI::CommonAPIEventCallback
```
Erfan-Ahmadi: 0. Let's get rid of `ICommonAPIEventCallback` because it's redundant and used in only 1 place to introduce `setLogger`
1. let's rename this class to `EventHandler` because we might decide to handle other events in the example with it.

With our current event handling design we need an instance of the event handler in order to get the events, the default one worked well until we needed to handle some events such as resize on our own.

Now the problem statement is "How can the event handler access the members of our example class" i.e the relation between the two.

## Current Solution
the current ugly solution right now is the `OnResize` class and the example inheriting from it to implement the virtual function.
In this case, What would we do if we wanted to specialize some other events like `onWindowRotated` or `onWindowMaximized`? Do we create new classes and inherit from them? Or do we have a single class `OnEvents` and have all the functions there and example class inherits from `OnEvents`? (may be actually a solution but not a good one)
We are in need of a solution that connects event handling to examples.

## Solution 1 - examples inherit `CommonAPIEventHandler`
Let the examples that use window inherit from `CommonAPIEventHandler` and then user passes the event handler itself in `InitParams` instead of letting CommonAPI construct it for them.
this way user can override the callbacks it needs.

I'm not so sure about this one because it kinda has dependancy on logger and input system which common apu provides for us, but we can simply have `setter` functins for them.

IF we make it work This solution would be convenient and better than the currentnt one and we wouldn't even need to have ~~Resize~~EventHandler 

## Solution 2 - register std::function
Our examples don't inherit from anything but ~~Resize~~EventHandler contains a bunch of `std::function`s with correct signatures like `std::function<void(uint32_t,uint32_t)>`

unless there is some weird reason not to use a std::function this is a pretty good choice.

## Solution 3 - the event functions in ApplicationFramework

This solution has the GraphicalApplication (or maybe IGraphicalApplicationFramework) define virtual functions like `OnResize`, and the `CommonAPI::CommonAPIEventCallback` stores a `IGraphicalApplicationFramework*` and has a function like `setIGraphicalApplication` and we can so something like:

```c++
bool onWindowResized_impl(uint32_t w, uint32_t h) override
{
	m_logger.log("Window resized to { %u, %u }", nbl::system::ILogger::ELL_DEBUG, w, h);
        if(graphicalApp)
           graphicalApp->onResize(w,h);
	return true;
}
```

I like this solution, but we're re-defining some function like `OnResize` similar to `onWindowResized_impl` again as opposed to Solution#1 where we straight up inherit from `CommonAPI::CommonAPIEventCallback` and the example will be it's own event handler.

the problem with Solution#1 is having to decide at compile time whether our example will need window and events but It's not a big deal.

We can combine this with Solution#1 and Have `GraphicalApplication inherit ICommonAPIEventCallback` or even one step further and `nbl::ui::IGraphicalApplicationFramework inherit nbl::ui::IWindow::IEventCallback`
It has the same downside of Solution#1 where we have to move the callback creation out of `CommonAPI::Init `and probably put it in `InitParams`

Let me know if you have other solutions in mind. we need to come to a conclusion about this.

05.NablaTutorialExample/main.cpp
```diff
@@ -60,7 +60,7 @@ class NablaTutorialExampleApp : public ApplicationBase
 	std::array<nbl::video::IGPUQueue*, CommonAPI::InitOutput::MaxQueuesCount> queues = { nullptr, nullptr, nullptr, nullptr };
 	nbl::core::smart_refctd_ptr<nbl::video::ISwapchain> swapchain;
 	nbl::core::smart_refctd_ptr<nbl::video::IGPURenderpass> renderpass;
-	std::array<nbl::core::smart_refctd_ptr<nbl::video::IGPUFramebuffer>, CommonAPI::InitOutput::MaxSwapChainImageCount> fbo;
+	nbl::core::smart_refctd_dynamic_array<nbl::core::smart_refctd_ptr<nbl::video::IGPUFramebuffer>> fbo;
```
Erfan-Ahmadi, line 63: Is there any reason why you didn't do this change for Example 02? 
Also What was the motivation to use dynamic array instead of std::array with MaxSwapChainImages?

Erfan-Ahmadi: nvm I figured it out

14.ComputeScan/main.cpp
```diff
@@ -20,16 +20,19 @@ class ComputeScanApp : public NonGraphicalApplicationBase
...
+		auto initOutput = CommonAPI::Init(std::move(initParams));
+
 		system = std::move(initOutput.system);
 		auto gl = std::move(initOutput.apiConnection);
 		auto logger = std::move(initOutput.logger);
 		auto gpuPhysicalDevice = std::move(initOutput.physicalDevice);
 		auto logicalDevice = std::move(initOutput.logicalDevice);
 		auto queues = std::move(initOutput.queues);
-		auto renderpass = std::move(initOutput.renderpass);
+		auto renderpass = std::move(initOutput.renderToSwapchainRenderpass);
```
devshgraphicsprogramming, line 35: this should have been not used at all

55.RGB18E7S3/main.cpp
```diff
@@ -43,15 +43,19 @@ int main()
...
+    auto initOutput = CommonAPI::Init(std::move(initParams));
+
 	auto system = std::move(initOutput.system);
     auto gl = std::move(initOutput.apiConnection);
     auto logger = std::move(initOutput.logger);
     auto gpuPhysicalDevice = std::move(initOutput.physicalDevice);
     auto logicalDevice = std::move(initOutput.logicalDevice);
     auto queues = std::move(initOutput.queues);
-    auto renderpass = std::move(initOutput.renderpass);
+    auto renderpass = std::move(initOutput.renderToSwapchainRenderpass);
```
devshgraphicsprogramming, line 58: this example is headless

common/CommonAPI.h
```diff
@@ -204,90 +204,19 @@ class CommonAPI
...
 		{
-			m_logger.log("Window closed");
-			return true;
+			m_app = std::move(app);
 		}
+	protected:
+		nbl::ui::IGraphicalApplicationFramework* m_app = nullptr;
 	};
+
 	class CommonAPIEventCallback : public ICommonAPIEventCallback
```
devshgraphicsprogramming, line 220: why do you need this weird inheritance?

common/CommonAPI.h
```diff
@@ -867,851 +485,152 @@ class CommonAPI
...
+
+
+	class IRetiredSwapchainResources: public nbl::video::ICleanup
 	{
-		SFeatureRequest<nbl::video::IAPIConnection::E_FEATURE> requiredInstanceFeatures = {};
-		SFeatureRequest<nbl::video::IAPIConnection::E_FEATURE> optionalInstanceFeatures = {};
-		SFeatureRequest<nbl::video::ILogicalDevice::E_FEATURE> requiredDeviceFeatures = {};
-		SFeatureRequest<nbl::video::ILogicalDevice::E_FEATURE> optionalDeviceFeatures = {};
+	public:
+		// nbl::core::smart_refctd_ptr<nbl::video::ISwapchain> oldSwapchain = nullptr; // this gets dropped along with the images
```
devshgraphicsprogramming, line 572: ok then fully remove

common/CommonAPI.h
```diff
@@ -1818,6 +736,9 @@ class GraphicalApplication : public nbl::system::IApplicationFramework, public n
 	void recreateSurface() override
 	{
 	}
+	void onResize(uint32_t w, uint32_t h) override
+	{
+	}
```
devshgraphicsprogramming, line 741: no, you CANNOT HAVE THIS!

common/CommonAPI.h
Erfan-Ahmadi: I suggested to remove this in a PR comment as well

55.RGB18E7S3/main.cpp
Erfan-Ahmadi: same as example 02 where an old fbo signature is seen

PR #12, opened 2022-08-10
Resizing maintenance chapter 2

deprilula28:
## Descriptions

Adresses https://github.com/Devsh-Graphics-Programming/Nabla/issues/406 and implements a blit-based solution for https://github.com/Devsh-Graphics-Programming/Nabla/issues/409

## TODO
- [ ] PR Reviews
- [ ] TODOs in the code: 
    - [ ] https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/blob/c3b8a7c563aef94ad460ce7cb6c1a47c88d1d65b/02.ComputeShader/main.cpp#L452
    - [ ] https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/blob/c3b8a7c563aef94ad460ce7cb6c1a47c88d1d65b/02.ComputeShader/main.cpp#L350
    - [ ] https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/blob/c3b8a7c563aef94ad460ce7cb6c1a47c88d1d65b/02.ComputeShader/main.cpp#L533
    - [ ] https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/blob/c3b8a7c563aef94ad460ce7cb6c1a47c88d1d65b/common/CommonAPI.h#L887
- [ ] Verify that other examples still compile & etc

### DIFF
--- a/02.ComputeShader/main.cpp
+++ b/02.ComputeShader/main.cpp
@@ -35,6 +35,9 @@ class ComputeShaderSampleApp : public ApplicationBase
 
 	int32_t m_resourceIx = -1;
 
+	std::array<core::smart_refctd_ptr<video::IGPUDescriptorSet>, 2> m_outputTargetDescriptorSet;
+	std::array<core::smart_refctd_ptr<video::IGPUImageView>, 2> m_outputTargetImageView;
+
 	core::smart_refctd_ptr<video::IGPUSemaphore> m_imageAcquire[FRAMES_IN_FLIGHT] = { nullptr };
 	core::smart_refctd_ptr<video::IGPUSemaphore> m_renderFinished[FRAMES_IN_FLIGHT] = { nullptr };
 	core::smart_refctd_ptr<video::IGPUFence> m_frameComplete[FRAMES_IN_FLIGHT] = { nullptr };
@@ -42,11 +45,8 @@ class ComputeShaderSampleApp : public ApplicationBase
 	core::smart_refctd_ptr<video::IGPUCommandBuffer> m_cmdbuf[FRAMES_IN_FLIGHT] = { nullptr };
 
 	core::smart_refctd_ptr<video::IGPUComputePipeline> m_pipeline = nullptr;
-	core::vector<core::smart_refctd_ptr<video::IGPUDescriptorSet>> m_descriptorSets;
 	core::smart_refctd_ptr<video::IGPUDescriptorSetLayout> m_descriptorSetLayout;
 
-	// These can be removed after descriptor lifetime tracking
-	core::vector<core::smart_refctd_ptr<video::IGPUImageView>> m_swapchainImageViews;
 	core::smart_refctd_ptr<video::IGPUImageView> m_inImageView;
 
 	core::deque<CommonAPI::IRetiredSwapchainResources*> m_qRetiredSwapchainResources;
@@ -56,9 +56,7 @@ class ComputeShaderSampleApp : public ApplicationBase
 
 	struct CSwapchainResources : public CommonAPI::IRetiredSwapchainResources
 	{
-		nbl::core::smart_refctd_ptr<nbl::video::IGPUImageView> oldImageView = nullptr;
 		nbl::core::smart_refctd_ptr<nbl::video::IGPUImage> oldImage = nullptr;
-		nbl::core::smart_refctd_ptr<nbl::video::IGPUDescriptorSet> descriptorSet = nullptr;
 
 		~CSwapchainResources() override {}
 	};
@@ -119,70 +117,6 @@ public:
 		auto& img = m_swapchainImages->begin()[i];
 		img = swapchain->createImage(i);
 		assert(img);
-		{
-			video::IGPUImageView::SCreationParams viewParams;
-			viewParams.format = img->getCreationParameters().format;
-			viewParams.viewType = asset::IImageView<video::IGPUImage>::ET_2D;
-			viewParams.subresourceRange.aspectMask = asset::IImage::EAF_COLOR_BIT;
-			viewParams.subresourceRange.baseMipLevel = 0u;
-			viewParams.subresourceRange.levelCount = 1u;
-			viewParams.subresourceRange.baseArrayLayer = 0u;
-			viewParams.subresourceRange.layerCount = 1u;
-			viewParams.image = img;
-
-			m_swapchainImageViews[i] = logicalDevice->createImageView(std::move(viewParams));
-			assert(m_swapchainImageViews[i]);
-		}
-
-		const uint32_t descriptorPoolSizeCount = 1u;
-		video::IDescriptorPool::SDescriptorPoolSize poolSizes[descriptorPoolSizeCount];
-		poolSizes[0].type = asset::EDT_STORAGE_IMAGE;
-		poolSizes[0].count = 2u;
-
-		video::IDescriptorPool::E_CREATE_FLAGS descriptorPoolFlags =
-			static_cast<video::IDescriptorPool::E_CREATE_FLAGS>(0);
-
-		core::smart_refctd_ptr<video::IDescriptorPool> descriptorPool
-			= logicalDevice->createDescriptorPool(descriptorPoolFlags, 1,
-				descriptorPoolSizeCount, poolSizes);
-
-		m_descriptorSets[i] = logicalDevice->createDescriptorSet(descriptorPool.get(),
-			core::smart_refctd_ptr(m_descriptorSetLayout));
-
-		const uint32_t writeDescriptorCount = 2u;
-
-		video::IGPUDescriptorSet::SDescriptorInfo descriptorInfos[writeDescriptorCount];
-		video::IGPUDescriptorSet::SWriteDescriptorSet writeDescriptorSets[writeDescriptorCount] = {};
-
-		// image2D -- my input
-		{
-			descriptorInfos[0].image.imageLayout = asset::IImage::EL_GENERAL;
-			descriptorInfos[0].image.sampler = nullptr;
-			descriptorInfos[0].desc = m_inImageView;
-
-			writeDescriptorSets[0].dstSet = m_descriptorSets[i].get();
-			writeDescriptorSets[0].binding = 1u;
-			writeDescriptorSets[0].arrayElement = 0u;
-			writeDescriptorSets[0].count = 1u;
-			writeDescriptorSets[0].descriptorType = asset::EDT_STORAGE_IMAGE;
-			writeDescriptorSets[0].info = &descriptorInfos[0];
-		}
-
-		// image2D -- swapchain image
-		{
-			descriptorInfos[1].image.imageLayout = asset::IImage::EL_GENERAL;
-			descriptorInfos[1].image.sampler = nullptr;
-			descriptorInfos[1].desc = m_swapchainImageViews[i]; // shouldn't IGPUDescriptorSet hold a reference to the resources in its descriptors?
-
-			writeDescriptorSets[1].dstSet = m_descriptorSets[i].get();
-			writeDescriptorSets[1].binding = 0u;
-			writeDescriptorSets[1].arrayElement = 0u;
-			writeDescriptorSets[1].count = 1u;
-			writeDescriptorSets[1].descriptorType = asset::EDT_STORAGE_IMAGE;
-			writeDescriptorSets[1].info = &descriptorInfos[1];
-		}
-
-		logicalDevice->updateDescriptorSets(writeDescriptorCount, writeDescriptorSets, 0u, nullptr);
 	}
 
 	void onAppInitialized_impl() override
@@ -239,7 +173,6 @@ public:
 
 		const uint32_t swapchainImageCount = swapchain->getImageCount();
 		m_swapchainImages = core::make_refctd_dynamic_array<core::smart_refctd_dynamic_array<core::smart_refctd_ptr<video::IGPUImage>>>(swapchainImageCount);
-		m_swapchainImageViews.resize(swapchainImageCount);
 
 		video::IGPUObjectFromAssetConverter CPU2GPU;
 
@@ -378,7 +311,6 @@ public:
 		}
 		assert(m_inImageView);
 
-		m_descriptorSets.resize(swapchainImageCount);
 		for (uint32_t i = 0u; i < swapchainImageCount; ++i)
 		{
 			createSwapchainImage(i);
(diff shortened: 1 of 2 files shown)


common/CommonAPI.h
```diff
@@ -722,23 +711,281 @@ class CommonAPI
 };
 
 #ifndef _NBL_PLATFORM_ANDROID_
-class GraphicalApplication : public nbl::system::IApplicationFramework, public nbl::ui::IGraphicalApplicationFramework
+class GraphicalApplication : public CommonAPI::CommonAPIEventCallback, public nbl::system::IApplicationFramework, public nbl::ui::IGraphicalApplicationFramework
```
devshgraphicsprogramming, line 714: what's the reason for this inheritance!?

common/CommonAPI.h
```diff
@@ -722,23 +711,281 @@ class CommonAPI
...
 {
 protected:
 	~GraphicalApplication() {}
+
+	static constexpr uint64_t MAX_TIMEOUT = 99999999999999ull;
```
devshgraphicsprogramming, line 719: please dont put it globally here

common/CommonAPI.h
```diff
@@ -722,23 +711,281 @@ class CommonAPI
...
+
+	uint32_t m_frameIx = 0;
+	nbl::core::deque<CommonAPI::IRetiredSwapchainResources*> m_qRetiredSwapchainResources;
```
devshgraphicsprogramming, line 722: shhouldn't it be a `deque<unique_ptr<CommonAPI::IRetiredSwapchainResources>>` !?

otherwise your created object will die at line 738 when it goes out of scope!

common/CommonAPI.h
```diff
@@ -722,23 +711,281 @@ class CommonAPI
...
+
+	bool tryAcquireImage(nbl::video::ISwapchain* swapchain, nbl::video::IGPUSemaphore* waitSemaphore, uint32_t* imgnum)
+	{
+		if (swapchain->acquireNextImage(MAX_TIMEOUT, waitSemaphore, nullptr, imgnum) == nbl::video::ISwapchain::EAIR_SUCCESS)
+		{
+			if (m_swapchainIteration > m_imageSwapchainIterations[*imgnum])
+			{
+				auto retiredResources = onCreateResourcesWithSwapchain(*imgnum).release();
+				m_imageSwapchainIterations[*imgnum] = m_swapchainIteration;
+				if (retiredResources) CommonAPI::retireSwapchainResources(m_qRetiredSwapchainResources, retiredResources);
```
devshgraphicsprogramming, line 738: `onCreateResourcesWithSwapchain` returns a `unique_ptr` which will kill the object when `retiredResources` goes out of scope (unless the unique ptr gets moved).

so the signature of `retireSwapchainResources` needs to take the retired resources as a `unique_ptr<>&&` and it needs to get moved, and the type of the deque needs to change to use `unique_ptr`

common/CommonAPI.h
```diff
@@ -722,23 +711,281 @@ class CommonAPI
...
+
+#ifdef ANTI_FLICKER
+	std::array<nbl::core::smart_refctd_ptr<nbl::video::IGPUImage>, 2> m_tripleBufferRenderTargets;
+
+	struct PresentedFrameInfo
+	{
+		uint64_t width : 14;
+		uint64_t height : 14;
+		uint64_t resourceIx : 8; // (Frame in flight)
+		uint64_t frameIx : 28; // (Total amount of frames rendered so far / frame index)
```
devshgraphicsprogramming, line 754: assuming you get 9000 FPS, this will overflow after 8 hours of uptime, will something break?

common/CommonAPI.h
```diff
@@ -722,23 +711,281 @@ class CommonAPI
...
-	void recreateSurface() override
+	) : nbl::system::IApplicationFramework(_localInputCWD, _localOutputCWD, _sharedInputCWD, _sharedOutputCWD),
+		CommonAPI::CommonAPIEventCallback(nullptr, nullptr),
+		m_qRetiredSwapchainResources(),
+		m_imageSwapchainIterations{}
+	{}
+
+	std::unique_lock<std::mutex> recreateSwapchain(
+		uint32_t w, uint32_t h, 
+		nbl::video::ISwapchain::SCreationParams& swapchainCreationParams,
```
devshgraphicsprogramming, line 828: shouldn't the creation parameters be persistent (immutable) and therefore not an argument to this method?

common/CommonAPI.h
```diff
@@ -515,9 +501,13 @@ class CommonAPI
 
 		if (!headlessCompute)
 		{
+			if (!params.windowCb)
+			{
+				params.windowCb = nbl::core::make_smart_refctd_ptr<EventCallback>(nbl::core::smart_refctd_ptr(result.inputSystem), system::logger_opt_smart_ptr(nbl::core::smart_refctd_ptr(result.logger)));
+			}
```
devshgraphicsprogramming, line 507: explain this?

common/CommonAPI.h
```diff
@@ -722,23 +711,281 @@ class CommonAPI
...
+				m_imageSwapchainIterations[*imgnum] = m_swapchainIteration;
+				if (retiredResources) CommonAPI::retireSwapchainResources(m_qRetiredSwapchainResources, retiredResources);
+			}
+
+			return true;
+		}
+		return false;
+	}
+
+#ifdef ANTI_FLICKER
```
devshgraphicsprogramming, line 746: do not use the preprocessor like this, use `template<bool tripleBufferedSwapchain` on the class definition like god intended

02.ComputeShader/main.cpp
```diff
@@ -1,14 +1,13 @@
 #define _NBL_STATIC_LIB_
 #include <nabla.h>
 
+#define ANTI_FLICKER
```
devshgraphicsprogramming, line 4: this is a nope

common/CommonAPI.h
devshgraphicsprogramming: its not like you're gonna change formats, min image count etc.

Perhaps we should move `CommonAPI::createSwapchain` method to `IGraphcalApplication`.

Then only ask for things _that can feasibly change_ in `recreateSwapchain` which are:
- width
- height
- preTransform?

Add a `ISwapchain::getCreationParams()`, patch them up with changes, and use that.

common/CommonAPI.h
```diff
@@ -722,23 +711,281 @@ class CommonAPI
...
+
+	// Returns retired resources
+	virtual std::unique_ptr<CommonAPI::IRetiredSwapchainResources> onCreateResourcesWithSwapchain(const uint32_t imageIndex) { return nullptr; }
+
+	bool tryAcquireImage(nbl::video::ISwapchain* swapchain, nbl::video::IGPUSemaphore* waitSemaphore, uint32_t* imgnum)
+	{
+		if (swapchain->acquireNextImage(MAX_TIMEOUT, waitSemaphore, nullptr, imgnum) == nbl::video::ISwapchain::EAIR_SUCCESS)
+		{
+			if (m_swapchainIteration > m_imageSwapchainIterations[*imgnum])
+			{
```
devshgraphicsprogramming, line 735: it makes no sense for the `swapchain` to be an argument while you're hardcoding to use `m_imageSwapchainIterations` and retiring to an implicit queue

either have everything implicit, OR take everything as an argument

common/CommonAPI.h
```diff
@@ -722,23 +711,281 @@ class CommonAPI
...
 protected:
 	~GraphicalApplication() {}
+
+	static constexpr uint64_t MAX_TIMEOUT = 99999999999999ull;
+
+	uint32_t m_frameIx = 0;
+	nbl::core::deque<CommonAPI::IRetiredSwapchainResources*> m_qRetiredSwapchainResources;
+	uint32_t m_swapchainIteration = 0;
+	std::array<uint32_t, CommonAPI::InitOutput::MaxSwapChainImageCount> m_imageSwapchainIterations;
+	std::mutex m_swapchainPtrMutex;
```
devshgraphicsprogramming, line 725: can we wrap this up into a small struct/class and make `tryAcquireImage` a method of it?

02.ComputeShader/main.cpp
devshgraphicsprogramming: use a template bool on the `ApplicationBase`

02.ComputeShader/main.cpp
```diff
@@ -124,102 +115,26 @@ class ComputeShaderSampleApp : public ApplicationBase
...
 	}
 
 	void onAppInitialized_impl() override
 	{
-		const auto swapchainImageUsage = static_cast<asset::IImage::E_USAGE_FLAGS>(asset::IImage::EUF_COLOR_ATTACHMENT_BIT | asset::IImage::EUF_STORAGE_BIT);
+		const auto swapchainImageUsage = static_cast<asset::IImage::E_USAGE_FLAGS>(asset::IImage::EUF_COLOR_ATTACHMENT_BIT | asset::IImage::EUF_STORAGE_BIT | asset::IImage::EUF_TRANSFER_DST_BIT);
 		std::array<asset::E_FORMAT, 1> acceptableSurfaceFormats = { asset::EF_B8G8R8A8_UNORM };
 
 		CommonAPI::InitParams initParams;
+		initParams.windowCb = core::smart_refctd_ptr<CommonAPI::CommonAPIEventCallback>(this);
```
devshgraphicsprogramming, line 126: what on earth is going on here?

DO NOT MAKE THE APPLICATION THE CALLBACK!

02.ComputeShader/main.cpp
```diff
@@ -418,20 +331,108 @@ class ComputeShaderSampleApp : public ApplicationBase
 		}
 	}
 
-	void onAppTerminated_impl() override
+	std::unique_ptr<CommonAPI::IRetiredSwapchainResources> onCreateResourcesWithSwapchain(const uint32_t imgnum)
 	{
-		logicalDevice->waitIdle();
+		logger->log("onCreateResourcesWithSwapchain(%i)", system::ILogger::ELL_INFO, imgnum);
+		CSwapchainResources* retiredResources(new CSwapchainResources{});
```
devshgraphicsprogramming, line 337: know about `make_unique` ?

common/CommonAPI.h
```diff
@@ -722,23 +711,281 @@ class CommonAPI
...
+
+	// Returns retired resources
+	virtual std::unique_ptr<CommonAPI::IRetiredSwapchainResources> onCreateResourcesWithSwapchain(const uint32_t imageIndex) { return nullptr; }
```
devshgraphicsprogramming, line 728: this should probably take a `const CResizableSwapchainUtils* swapchain` in case we make an example with multiple resizable swapchains in the future (to tell apart which swapchain we want to deal with)

devshgraphicsprogramming: also to automatically set `retiredResources->retiredFrameId = swapchainUtils->m_frameIx;` on the returned/created item

(then we make this an `inline` and make a `virtual unique_ptr<> onCreateResourcesWithSwapchain_impl`)

common/CommonAPI.h
devshgraphicsprogramming: maybe make this a method of `CResizableSwapchainUtils`

common/CommonAPI.h
```diff
@@ -722,23 +711,281 @@ class CommonAPI
...
 	{
+		auto logicalDevice = getLogicalDevice();
+		if (fence)
+		{
+			logicalDevice->blockForFences(1u, &fence.get());
+			if (m_frameIx >= framesInFlight) CommonAPI::dropRetiredSwapchainResources(m_qRetiredSwapchainResources, m_frameIx - framesInFlight);
+		}
+		else
+			fence = logicalDevice->createFence(static_cast<nbl::video::IGPUFence::E_CREATE_FLAGS>(0));
 	}
```
devshgraphicsprogramming, line 856: due to implicit dependency on `m_frameIx` make it a method of `CResizableSwapchainUtils`

common/CommonAPI.h
```diff
@@ -722,23 +711,281 @@ class CommonAPI
...
+
+	void waitForFrame(
+		uint32_t framesInFlight,
+		nbl::core::smart_refctd_ptr<nbl::video::IGPUFence>& fence)
 	{
+		auto logicalDevice = getLogicalDevice();
+		if (fence)
+		{
+			logicalDevice->blockForFences(1u, &fence.get());
+			if (m_frameIx >= framesInFlight) CommonAPI::dropRetiredSwapchainResources(m_qRetiredSwapchainResources, m_frameIx - framesInFlight);
```
devshgraphicsprogramming, line 852: can we somehow deduce `framesInFlight` ?

devshgraphicsprogramming: so user need not provide it?

02.ComputeShader/main.cpp
```diff
@@ -443,30 +444,17 @@ class ComputeShaderSampleApp : public ApplicationBase
...
-			if (m_frameIx >= FRAMES_IN_FLIGHT) CommonAPI::dropRetiredSwapchainResources(m_qRetiredSwapchainResources, m_frameIx - FRAMES_IN_FLIGHT);
-		}
-		else
-			fence = logicalDevice->createFence(static_cast<video::IGPUFence::E_CREATE_FLAGS>(0));
 
-		// acquire image 
-		uint32_t imgnum = 0u;
-		std::unique_lock guard = acquireImage(&imgnum);
+		waitForFrame(FRAMES_IN_FLIGHT, fence);
+		logicalDevice->resetFences(1u, &fence.get());
```
devshgraphicsprogramming, line 449: why not make the `waitForFrame` reset the fence after blocking? (fold it into itself)

common/CommonAPI.h
devshgraphicsprogramming: also make the all the members private :D

common/CommonAPI.h
```diff
@@ -722,23 +711,281 @@ class CommonAPI
...
+		while (true)
+		{
+			std::unique_lock guard(m_swapchainPtrMutex);
+			swapchain = swapchainRef;
+			if (tryAcquireImage(swapchain.get(), waitSemaphore, imgnum))
+			{
+				return swapchain;
+			}
+		}
+	}
```
devshgraphicsprogramming, line 873: thinking that `acquire` should be a public method of `CResizableSwapchainUtils`

common/CommonAPI.h
```diff
@@ -722,23 +711,281 @@ class CommonAPI
...
+
+#ifdef ANTI_FLICKER
+	void immediateImagePresent(nbl::video::IGPUQueue* queue, nbl::video::ISwapchain* swapchain, nbl::core::smart_refctd_ptr<nbl::video::IGPUImage>* swapchainImages, uint32_t frameIx, uint32_t lastRenderW, uint32_t lastRenderH)
```
devshgraphicsprogramming, line 876: could we roll this into `recreateSwapchain` ?

I.e. have an override of `recreateSwapchain` which takes just the extra arguments of `video::IGPUQueue* presentQueue, const uint32_t ` and that would be the one that does the additional triple buffer intermediate blit

common/CommonAPI.h
```diff
@@ -722,23 +711,281 @@ class CommonAPI
...
+				if (retiredResources) CommonAPI::retireSwapchainResources(m_qRetiredSwapchainResources, retiredResources);
+			}
+
+			return true;
+		}
+		return false;
+	}
+
+#ifdef ANTI_FLICKER
+	std::array<nbl::core::smart_refctd_ptr<nbl::video::IGPUImage>, 2> m_tripleBufferRenderTargets;
```
devshgraphicsprogramming, line 747: roll this into `CResizableSwapchainUtils<bool tripleBuffered>`

devshgraphicsprogramming: you need a class specialization  or something to make the presence of the member conditional

devshgraphicsprogramming:
This has turned into nasty sphagetti, I have a feeling I need to rewrite parts of it myself.

PR #17, opened 2022-09-08
Check surface capabilities to fix debug layer errors

deprilula28:
On some of the images for example 09, the resize fails, leading to debug layer errors and the image not being presented. This fixes the issue by querying the surface capabilities for min & max extents.


Erfan-Ahmadi:
Problem origin was windows not resizing the surface to the requested resolution but the whole window

PR #24, opened 2022-12-31
Add initial Schussler test

sunho:
Add initial Schussler test. Code needs to be polished a bit but functionalities are there.

### DIFF
--- a/62.SchusslerTest/main.cpp
+++ b/62.SchusslerTest/main.cpp
@@ -154,26 +154,6 @@ public:
 
   APP_CONSTRUCTOR(SchusslerTestApp)
 
-  	#include "nbl/nblpack.h"
-
-    struct SphereVertex
-		{
-			std::array<float,3> pos;
-			std::array<uint8_t,4> color;
-			std::array<float,2> uv;
-			CQuantNormalCache::value_type_t<EF_A2B10G10R10_SNORM_PACK32> normal;
-		} PACK_STRUCT;
-
-		struct VertexWithGNormal {
-    std::array<float, 3> pos;
-    std::array<uint8_t, 4> color;
-    std::array<float, 2> uv;
-    CQuantNormalCache::value_type_t<EF_A2B10G10R10_SNORM_PACK32> normal;
-    std::array<float, 3> gnormal;
-  } PACK_STRUCT;
-  
-	#include "nbl/nblunpack.h"
-
   void onAppInitialized_impl() override {
     const auto swapchainImageUsage = static_cast<asset::IImage::E_USAGE_FLAGS>(
         asset::IImage::EUF_COLOR_ATTACHMENT_BIT);
@@ -219,56 +199,8 @@ public:
                ->getQuantNormalCache();
     constexpr uint32_t INSTANCE_COUNT = 25u;
 
-    auto geometryObject = geometryCreator->createSphereMesh(0.5f, 8u, 8u);
-    constexpr size_t vertexSize = sizeof(VertexWithGNormal);
-
-    // Change vertex buffer to add gemoetric normal
-    geometryObject.inputParams = {
-        0b11111u,
-        0b1u,
-        {
-            {0u, EF_R32G32B32_SFLOAT, offsetof(VertexWithGNormal, pos)},
-            {0u, EF_R8G8B8A8_UNORM, offsetof(VertexWithGNormal, color)},
-            {0u, EF_R32G32_SFLOAT, offsetof(VertexWithGNormal, uv)},
-            {0u, EF_A2B10G10R10_SNORM_PACK32,
-             offsetof(VertexWithGNormal, normal)},
-            {0u, EF_R32G32B32_SFLOAT,
-             offsetof(VertexWithGNormal, gnormal)},
-        },
-        {vertexSize, EVIR_PER_VERTEX}};
-    uint32_t *indices =
-        reinterpret_cast<uint32_t *>(geometryObject.indexBuffer.buffer->getPointer());
-    auto oldVertBufferPtr = geometryObject.bindings[0].buffer;
-    SphereVertex *oldVertBuffer =
-        reinterpret_cast<SphereVertex *>(oldVertBufferPtr->getPointer());
-    auto vtxBuf =
-        core::make_smart_refctd_ptr<asset::ICPUBuffer>(oldVertBufferPtr->getSize() / sizeof(SphereVertex) * sizeof(VertexWithGNormal));
-    VertexWithGNormal *newVertBuffer =
-        reinterpret_cast<VertexWithGNormal*>(vtxBuf->getPointer());
-
-    for (size_t i = 0; i < geometryObject.indexCount; i += 3) {
-      uint32_t vi0 = indices[i];
-      uint32_t vi1 = indices[i + 1];
-      uint32_t vi2 = indices[i + 2];
-      auto &v0 = oldVertBuffer[vi0];
-      auto &v1 = oldVertBuffer[vi1];
-      auto &v2 = oldVertBuffer[vi2];
-      core::vectorSIMDf pos0(v0.pos.data());
-      core::vectorSIMDf pos1(v1.pos.data());
-      core::vectorSIMDf pos2(v2.pos.data());
-      core::vectorSIMDf newNormal = core::cross(pos2 - pos0, pos1 - pos0);
-      std::array<float,3> newNormalPacked = {newNormal[0], newNormal[1], newNormal[2]};
-      newVertBuffer[vi0] = {v0.pos, v0.color, v0.uv, v0.normal,
-                               newNormalPacked};
-      newVertBuffer[vi1] = {v1.pos, v1.color, v1.uv, v1.normal,
-                               newNormalPacked};
-      newVertBuffer[vi2] = {v2.pos, v2.color, v2.uv, v2.normal,
-                               newNormalPacked};
-    }
-    vtxBuf->setUsageFlags(vtxBuf->getUsageFlags() |
-                          asset::IBuffer::EUF_VERTEX_BUFFER_BIT);
-    geometryObject.bindings[0] = {0ull, std::move(vtxBuf)};
-
+    auto geometryObject = geometryCreator->createIcoSphere(0.5f, 2, true);
+    
     asset::SPushConstantRange rng[2];
     rng[0].offset = 0u;
     rng[0].size = sizeof(SPushConsts::vertStage);
--- a/62.SchusslerTest/main.cpp
+++ b/62.SchusslerTest/main.cpp
@@ -1,0 +1,527 @@
+// Copyright (C) 2018-2022 - DevSH Graphics Programming Sp. z O.O.
+// This file is part of the "Nabla Engine".
+// For conditions of distribution and use, see copyright notice in nabla.h
+
+#define _NBL_STATIC_LIB_
+#include <cstdio>
+#include <iostream>
+#include <nabla.h>
+
+#include "nbl/asset/utils/CGeometryCreator.h"
+#include "../common/Camera.hpp"
+#include "../common/CommonAPI.h"
+
+using namespace nbl;
+using namespace asset;
+using namespace video;
+using namespace core;
+
+/*
+        Uncomment for more detailed logging
+*/
+
+// #define NBL_MORE_LOGS
+
+class SchusslerTestApp : public ApplicationBase {
+  constexpr static uint32_t WIN_W = 1280;
+  constexpr static uint32_t WIN_H = 720;
+  constexpr static uint32_t SC_IMG_COUNT = 3u;
+  constexpr static uint32_t FRAMES_IN_FLIGHT = 5u;
+
+  static_assert(FRAMES_IN_FLIGHT > SC_IMG_COUNT);
+
+public:
+  nbl::core::smart_refctd_ptr<nbl::ui::IWindowManager> windowManager;
+  nbl::core::smart_refctd_ptr<nbl::ui::IWindow> window;
+  nbl::core::smart_refctd_ptr<CommonAPI::CommonAPIEventCallback> windowCb;
+  nbl::core::smart_refctd_ptr<nbl::video::IAPIConnection> apiConnection;
+  nbl::core::smart_refctd_ptr<nbl::video::ISurface> surface;
+  nbl::core::smart_refctd_ptr<nbl::video::IUtilities> utilities;
+  nbl::core::smart_refctd_ptr<nbl::video::ILogicalDevice> logicalDevice;
+  nbl::video::IPhysicalDevice *physicalDevice;
+  std::array<nbl::video::IGPUQueue *, CommonAPI::InitOutput::MaxQueuesCount>
+      queues = {nullptr, nullptr, nullptr, nullptr};
+  nbl::core::smart_refctd_ptr<nbl::video::ISwapchain> swapchain;
+  nbl::core::smart_refctd_ptr<nbl::video::IGPURenderpass> renderpass;
+  nbl::core::smart_refctd_dynamic_array<
+      nbl::core::smart_refctd_ptr<nbl::video::IGPUFramebuffer>>
+      fbo;
+  std::array<
+      std::array<nbl::core::smart_refctd_ptr<nbl::video::IGPUCommandPool>,
+                 CommonAPI::InitOutput::MaxFramesInFlight>,
+      CommonAPI::InitOutput::MaxQueuesCount>
+      commandPools; // TODO: Multibuffer and reset the commandpools
+  nbl::core::smart_refctd_ptr<nbl::system::ISystem> system;
+  nbl::core::smart_refctd_ptr<nbl::asset::IAssetManager> assetManager;
+  nbl::video::IGPUObjectFromAssetConverter::SParams cpu2gpuParams;
+  nbl::core::smart_refctd_ptr<nbl::system::ILogger> logger;
+  nbl::core::smart_refctd_ptr<CommonAPI::InputSystem> inputSystem;
+
+  nbl::video::IGPUObjectFromAssetConverter cpu2gpu;
+
+  core::smart_refctd_ptr<video::IGPUMeshBuffer> gpuMeshBuffer;
+  core::smart_refctd_ptr<IGPURenderpassIndependentPipeline>
+      gpuRenderpassIndependentPipeline;
+  core::smart_refctd_ptr<IGPUBuffer> gpuubo;
+  core::smart_refctd_ptr<IGPUDescriptorSet> gpuDescriptorSet1;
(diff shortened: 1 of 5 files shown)


62.SchusslerTest/main.cpp
```diff
@@ -0,0 +1,527 @@
...
+			CQuantNormalCache::value_type_t<EF_A2B10G10R10_SNORM_PACK32> normal;
+		} PACK_STRUCT;
+
+		struct VertexWithGNormal {
+    std::array<float, 3> pos;
+    std::array<uint8_t, 4> color;
+    std::array<float, 2> uv;
+    CQuantNormalCache::value_type_t<EF_A2B10G10R10_SNORM_PACK32> normal;
+    std::array<float, 3> gnormal;
+  } PACK_STRUCT;
```
devshgraphicsprogramming, line 173: you don't need this, its possible to retrieve from per-pixel derivatives

62.SchusslerTest/main.cpp
```diff
@@ -0,0 +1,527 @@
...
+      core::vectorSIMDf pos2(v2.pos.data());
+      core::vectorSIMDf newNormal = core::cross(pos2 - pos0, pos1 - pos0);
+      std::array<float,3> newNormalPacked = {newNormal[0], newNormal[1], newNormal[2]};
+      newVertBuffer[vi0] = {v0.pos, v0.color, v0.uv, v0.normal,
+                               newNormalPacked};
+      newVertBuffer[vi1] = {v1.pos, v1.color, v1.uv, v1.normal,
+                               newNormalPacked};
+      newVertBuffer[vi2] = {v2.pos, v2.color, v2.uv, v2.normal,
+                               newNormalPacked};
+    }
```
devshgraphicsprogramming, line 267: I don't get the point of doing all that.

62.SchusslerTest/shader.vert
```diff
@@ -0,0 +1,44 @@
+// Copyright (C) 2018-2020 - DevSH Graphics Programming Sp. z O.O.
+// This file is part of the "Nabla Engine".
+// For conditions of distribution and use, see copyright notice in nabla.h
+
+#version 430 core
+
+layout (location = 0) in vec3 Pos;
+layout (location = 3) in vec3 Normal;
```
devshgraphicsprogramming, line 8: you need UVs for the tangent for anisotropic BRDF

62.SchusslerTest/shader.vert
```diff
@@ -0,0 +1,44 @@
...
+    vec3 pos = to_right_hand(Pos);
+    outPos = pos;
+    vec4 projPos = pc.VP*vec4(pos, 1.0);
+    gl_Position = pc.VP*vec4(pos, 1.0) + vec4(gridPos*projPos.w,0.0,0.0);
+    outNormal = to_right_hand(normalize(Normal));
+    outLightPos = vec3(0.75,0.75,1.75);
+    outCamPos = vec3(0.0,0.0,6.5);
+    outGNormal = to_right_hand(normalize(Normal));
+    outAlpha = gridX * 0.2 + 0.1;
+    outAlphaY = gridY * 0.2;
```
devshgraphicsprogramming, line 43: if would be better to send over the `gridX,gridY` over to pixel shader and work it out there

62.SchusslerTest/shader.frag
```diff
@@ -0,0 +1,127 @@
...
+{
+    const float Intensity = 20.0;
+
+    vec3 L = LightPos-Pos;
+    vec3 Lnorm = normalize(L);
+    const float THRESHOLD = AlphaY+0.5;
+    float a2 = Alpha;
+    float a2formeta = AlphaY;
+    
+    vec3 wi = normalize(CamPos - Pos);
```
devshgraphicsprogramming, line 101: this will act like a point light, we want directional (constant `wi`) to be able to make comparisons

62.SchusslerTest/shader.frag
```diff
@@ -0,0 +1,127 @@
...
+    }
+    return brdf;
+}
+
+void main()
+{
+    const float Intensity = 20.0;
+
+    vec3 L = LightPos-Pos;
+    vec3 Lnorm = normalize(L);
```
devshgraphicsprogramming, line 96: also want constant `wo`

62.SchusslerTest/shader.frag
```diff
@@ -0,0 +1,127 @@
...
+    vec3 brdf_cos_wi_wo = evalBRDF(wi,wo,Normal);
+    vec3 brdf_cos_wi_wdasho = evalBRDF(wi,wdasho,Normal);
+    vec3 brdf_cos_wdashi_wo = evalBRDF(wdashi,wo,Normal);
+    vec3 brdf;
+    if (deviation<THRESHOLD) 
+    {
+        brdf = lambdap(wi, wt, Normal, GNormal) * (
+            G1(wo, Normal, a2formeta) * brdf_cos_wi_wo +
+            G1(wo, wt, a2formeta) * (1.0 - G1(wdasho, Normal, a2formeta)) * brdf_cos_wi_wdasho
+        ) + lambdat(wi, wt, Normal, GNormal) * brdf_cos_wdashi_wo * G1(wo, Normal, a2formeta);
```
devshgraphicsprogramming, line 118: big bug `Normal` needs to be normalized before use, per-pixel interpolation foreshortens it

62.SchusslerTest/shader.frag
```diff
@@ -0,0 +1,127 @@
...
+// Copyright (C) 2018-2020 - DevSH Graphics Programming Sp. z O.O.
+// This file is part of the "Nabla Engine".
+// For conditions of distribution and use, see copyright notice in nabla.h
+
+
+layout (location = 0) in vec3 Normal;
+layout (location = 1) in vec3 Pos;
+layout (location = 2) flat in vec3 LightPos;
+layout (location = 3) flat in vec3 CamPos;
+layout (location = 4) flat in vec3 GNormal;
```
devshgraphicsprogramming, line 11: geometric normal can be obtained with
```glsl
    vec3 dFdxPos = dFdx( Pos );
    vec3 dFdyPos = dFdy( Pos );
vec3 GNormal = normalize( cross(dFdxPos,dFdyPos ));
```

62.SchusslerTest/shader.frag
```diff
@@ -0,0 +1,127 @@
...
+    return brdf;
+}
+
+void main()
+{
+    const float Intensity = 20.0;
+
+    vec3 L = LightPos-Pos;
+    vec3 Lnorm = normalize(L);
+    const float THRESHOLD = AlphaY+0.5;
```
devshgraphicsprogramming, line 97: threshold for the deviation should not depend on the roughness!

deviation is intrinsically tied to what % of the overall surface are is the tangent microfacet, as deviation->0 the surface area ->0 and therefore so does `Lambda_t`

62.SchusslerTest/shader.frag
```diff
@@ -0,0 +1,127 @@
...
+    } else if (pc.testNum == TEST_AS) {
+        float nx = nbl_glsl_alpha2_to_phong_exp(a2);
+        float aa = 1.0-Alpha;
+        float ny = nbl_glsl_alpha2_to_phong_exp(aa*aa);
+        brdf = nbl_glsl_blinn_phong_cos_eval(_sample, inter, cache, nx, ny, ior);
+    } else if (pc.testNum == TEST_OREN_NAYAR) {
+        brdf = albedo*nbl_glsl_oren_nayar_cos_eval(_sample, inter_, a2);
+    } else if (pc.testNum == TEST_LAMBERT) {
+        brdf = albedo*nbl_glsl_lambertian_cos_eval(_sample);
+    }
```
devshgraphicsprogramming, line 87: these are the only BRDFs that should be using roughnesses

62.SchusslerTest/shader.frag
```diff
@@ -0,0 +1,127 @@
...
+    const float THRESHOLD = AlphaY+0.5;
+    float a2 = Alpha;
+    float a2formeta = AlphaY;
+    
+    vec3 wi = normalize(CamPos - Pos);
+    vec3 wo = Lnorm;
+    const float deviation = dot(Normal,GNormal);
+    vec3 wt = -normalize(Normal-deviation*GNormal);
+    vec3 wdasho = normalize(reflect(-wo, wt));
+    vec3 wdashi = normalize(reflect(-wi, wt));
```
devshgraphicsprogramming, line 106: 1. if two inputs to `reflect` are unit length there's no need to `normalize` them
2. the `-` on the input correct, GLSL `reflect` is retarded, you can use the Nabla one https://asawicki.info/news_1301_reflect_and_refract_functions.html

62.SchusslerTest/shader.frag
```diff
@@ -0,0 +1,127 @@
...
+    const float Intensity = 20.0;
+
+    vec3 L = LightPos-Pos;
+    vec3 Lnorm = normalize(L);
+    const float THRESHOLD = AlphaY+0.5;
+    float a2 = Alpha;
+    float a2formeta = AlphaY;
+    
+    vec3 wi = normalize(CamPos - Pos);
+    vec3 wo = Lnorm;
```
devshgraphicsprogramming, line 102: you're missing a basic check at the very start to ensure you're not "seeing" or "lighting" geometrically invisible macrosurfaces
```glsl
// geometric DPs
const float NdotV = dot(GNormal,wi);
const float NdotL = dot(GNormal,wo);
if (NdotV<nbl_glsl_FLT_MIN || NdotL<nbl_glsl_FLT_MIN)
   outColor = vec3(0.0,0.0,0.0,1.0);
else
{
// impl
}
```

62.SchusslerTest/shader.frag
```diff
@@ -0,0 +1,127 @@
...
+    return max(0.0, dot(a,b));
+}
+
+float ap(vec3 wi, vec3 wp, vec3 wg) {
+    return pdot(wi, wp)/pdot(wp, wg);
+}
+
+float at(vec3 wi, vec3 wt, vec3 wp, vec3 wg) {
+    return pdot(wi, wt) * sqrt(1-(pdot(wp,wg)*pdot(wp,wg))) / pdot(wp,wg);
+}
```
devshgraphicsprogramming, line 45: you don't use these functions anywhere but for `lambda`, so can remove and have optimized variants

62.SchusslerTest/shader.frag
```diff
@@ -0,0 +1,127 @@
...
+
+float lambdap(vec3 wi, vec3 wt, vec3 wp, vec3 wg) {
+    return ap(wi, wp, wg) / (ap(wi,wp,wg) + at(wi,wt,wp,wg));
+}
+
+float lambdat(vec3 wi, vec3 wt, vec3 wp, vec3 wg) {
+    return at(wi,wt,wp,wg) / (ap(wi,wp,wg) + at(wi,wt,wp,wg));
+}
```
devshgraphicsprogramming, line 53: `at` and `ap` both have a denominator `pdot(wp,wg)` which actually happens to be the `deviation`

since its in all of them there's no point applying it.

62.SchusslerTest/shader.frag
devshgraphicsprogramming: also use `1.0` or `1.f` instead of 1, cause GLSL is a stickler for types

62.SchusslerTest/shader.frag
devshgraphicsprogramming: furthermore `(ap+at)*deviation` can be optimized to
```
pdot(wi,wp)+pdot(wi,wt)*sqrt(1.f-deviation*deviation)
```

but since
```
wt = normalize(wg*deviation-wp) = (wg*deviation-wp)/length(wg*deviation-wp) = (wg*deviation-wp)/sqrt(1.f-deviation*deviation)
```
we have
```
pdot(wi,wp)+pdot(wi,wg*deviation-wp) == pdot(wi,wp)+pdot(wi,unnormalized_wt)
```

devshgraphicsprogramming: this way everything is much more numerically robust because it will work even when `deviation=0.f` since we don't have divisions by 0 anymore (unless `wi` is below the geometrical hemisphere, which is when we should never even start to shade the surface in the first place)

62.SchusslerTest/shader.frag
devshgraphicsprogramming, line 87: here you should be checking that `inter.isotropic.NdotV` and `_sample.NdotL` are valid (same check like for geometric microfacet you perform for the perturbed one)

62.SchusslerTest/shader.frag
devshgraphicsprogramming: Also to make your life even more simpler `lambda_t = 1.f-lambda_p` :D

62.SchusslerTest/shader.frag
```diff
@@ -0,0 +1,127 @@
...
+
+float G1(vec3 wo, vec3 wp, float a2) {
+    float dotNV = pdot(wp,wo);
+    float denomC = sqrt(a2 + (1.0f - a2) * dotNV * dotNV) + dotNV;
+
+    return 2.0f * dotNV / denomC;
+}
```
devshgraphicsprogramming, line 60: your G1 is completely wrong, it should be
```glsl
// `max` means its already after `max(,0.f)` treatment
// `X` means either L or V depending on whether you evaluate with `w_o` or `w_i
float schussler_lambda_den(float maxShadingNdotX, float maxSpecularTdotX)
{
   return maxShadingNdotX+maxSpecularTdotX;
}

// [0] lambda_p
// [1] lambda_t
vec2 lambda(float maxShadingNdotX, float lambda_den)
{
   const float lambda_p = maxShadingNdotX/lambda_den;
   return vec2(lambda_p,1.f-lambda_p);
}

// in Nabla, the half vector is denoted `H`, which is `w_m` in the Schussler paper
float G1(float HdotX, float GeomNdotX, float GeomNdotShadingN/*old deviation*/, float lambda_den)
{
   // microfacet cannot be facing away from the direction, which is why the Heaviside function was in Eq 13
   if (HdotX>0.f)
   {
      const float rcp_a_sum = GeomNdotShadingN/lambda_den;
      return min(GeomNdotX*rcp_a_sum,1.f);
   }
   return 0.f;
}
```

devshgraphicsprogramming: Your G2 becomes
```glsl
// even though we call `w_m` `H` its not a half vector, because no perfect specularity happens
float schussler_G2(float HdotV, float GeomNdotV, float NdotV, float HdotL, float GeomNdotL, float NdotL, float lambda_den)
{
   return schussler_G1(HdotV,GeomNdotV,NdotV,HdotL,GeomNDotL,NdotL,lambda_den);
}
```

62.SchusslerTest/shader.frag
devshgraphicsprogramming: basically this 

> Remember to implement pre `dot(V,geometricNormal)>FLT_MIN` and post `dot(L,geometricNormal)>FLT_MIN` geometrical feasibility checks.

The pre-post distrinction is only for importance sampling, if you know both up front (For eval or pdf computation) then you check both up front.

from https://github.com/Devsh-Graphics-Programming/Nabla/issues/310

62.SchusslerTest/shader.frag
devshgraphicsprogramming: In https://github.com/Devsh-Graphics-Programming/Nabla/issues/310 I call `lambda_p`, `shadingNormal_vpdf` instead.

The two definitions/derivations are compatible. This is the most optimal
```glsl
float nbl_glsl_bxdf_Schussler_shadingNormalVPDF(in float NdotX, in float clampedNdotX, in float GdotX, in float GdotN/*old deviation*/)
{
   assert(GeomNdotShadingN>=0.f);
   assert(XdotGeomN>0.f);
   return clampedNdotX/(clampedNdotX+max(NdotX-GdotN*GdotX,0.f));
}
```

We could define a
```glsl
// a_p/(a_p+a_t)
float nbl_glsl_bxdf_schussler_shadingNormal_vpdf(in float clampedNdotX, in float DenormTdotX)
{
   return NdotX/(NdotX+DenormTdotX);
}
```

devshgraphicsprogramming: Bonus points: for BSDFs we'd need to invert the sign of `NdotX` and `GdotX` whenever `GdotX<0.f`

Assuming that for BRDFs we check and skip all shading logic when `GdotX`, this means we can `abs` it in the above function

62.SchusslerTest/shader.frag
devshgraphicsprogramming: the G2 I quoted is wrong, should be
```glsl
float schussler_G2(float HdotV, float GdotV, float rcp_a_sumV, float HdotL, float GdotL, float rcp_a_sumL)
{
   return schussler_G1(HdotV,GdotV,rcp_a_sumV)*schussler_G1(HdotL,GdotL,rcp_a_sumL);
}
```
the `rcp_a_sum` is different for each direction L and V

devshgraphicsprogramming: I've proposed optimized variant called `shadingNormal_masking` in https://github.com/Devsh-Graphics-Programming/Nabla/issues/310 but it should be quadruple checked

62.SchusslerTest/shader.frag
devshgraphicsprogramming: Ok actually the `-` shouldn't be ther, because you want the relfected directions to point "away" from perturbed microfacet

62.SchusslerTest/shader.frag
```diff
@@ -91,37 +89,56 @@ vec3 evalBRDF(vec3 wi, vec3 wo, vec3 N) {
...
+    if (deviation < THRESHOLD) 
     {
-        brdf = lambdap(wi, wt, Normal, GNormal) * (
-            G1(wo, Normal, a2formeta) * brdf_cos_wi_wo +
-            G1(wo, wt, a2formeta) * (1.0 - G1(wdasho, Normal, a2formeta)) * brdf_cos_wi_wdasho
-        ) + lambdat(wi, wt, Normal, GNormal) * brdf_cos_wdashi_wo * G1(wo, Normal, a2formeta);
+        if (dot(wi, GN) >= 0.0) brdf += lambdap(wi, wt, N, GN) * G1(wo, N, wt, N, GN) * brdf_cos_wi_wo;
+        if (dot(wi, GN) >= 0.0) brdf += lambdap(wi, wt, N, GN) * G1(wo, wt,  wt, N, GN) 
+            * (1.0 - G1(wdasho, N,  wt, N, GN)) * brdf_cos_wi_wdasho;
+        if (dot(wdashi, GN) >= 0.0) brdf += lambdat(wi, wt, N, GN) * brdf_cos_wdashi_wo * G1(wo, N,  wt, N, GN);
```
devshgraphicsprogramming, line 129: the `if` statements are wrong, checks should be inside the `evalBRDF` functions (for now) and you need to check both `wi` and `wo` against shading normal, not geometrical normal!

PR #29, opened 2023-02-19
address compiling issues linux

pollend:
a set can't have a const type so this just changes the pointer to a non-const type. and added CWindowManagerX11 even though it doesn't work at the moment.

ref: https://github.com/Devsh-Graphics-Programming/Nabla/pull/454

https://stackoverflow.com/questions/6954906/does-c11-allow-vectorconst-t

### DIFF
--- a/common/CommonAPI.h
+++ b/common/CommonAPI.h
@@ -321,7 +321,7 @@ public:
 	{
 	public:
 		// ! this will get called after all physical devices go through filtering via `InitParams::physicalDeviceFilter`
-		virtual nbl::video::IPhysicalDevice* const selectPhysicalDevice(nbl::core::set<nbl::video::IPhysicalDevice* const> suitablePhysicalDevices) = 0;
+		virtual nbl::video::IPhysicalDevice* selectPhysicalDevice(nbl::core::set<nbl::video::IPhysicalDevice*> suitablePhysicalDevices) = 0;
 	};
 	
 	class CDefaultPhysicalDeviceSelector : public CommonAPI::IPhysicalDeviceSelector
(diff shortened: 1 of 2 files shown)


common/CommonAPI.h
```diff
@@ -321,7 +321,7 @@ class CommonAPI
 	{
 	public:
 		// ! this will get called after all physical devices go through filtering via `InitParams::physicalDeviceFilter`
-		virtual nbl::video::IPhysicalDevice* const selectPhysicalDevice(nbl::core::set<nbl::video::IPhysicalDevice* const> suitablePhysicalDevices) = 0;
+		virtual nbl::video::IPhysicalDevice* selectPhysicalDevice(nbl::core::set<nbl::video::IPhysicalDevice*> suitablePhysicalDevices) = 0;
```
devshgraphicsprogramming, line 324: ah yeah it should probably be `const core::set<T*>&` not `core::set<T* const>`

common/CommonAPI.h
```diff
@@ -495,8 +495,14 @@ class CommonAPI
 			params.windowCb->setInputSystem(nbl::core::smart_refctd_ptr(result.inputSystem));
 			if (!params.window)
 			{
-				result.windowManager = nbl::core::make_smart_refctd_ptr<nbl::ui::CWindowManagerWin32>(); // on the Android path
```
devshgraphicsprogramming, line 498: @sadiuk what was this comment about?

devshgraphicsprogramming: or @Crisspl 

I remember that Android was weird cause it fave you an already created window and there was nothing you could do about it + the initialization was split into several weird steps

PR #30, opened 2023-02-21
Path tracer features

achalpandeyy:
- [x] Straight to interactive
- [x] ~~Radeon Rays timeout~~
- [x] Reload


22.RaytracedAO/CommandLineHandler.hpp
```diff
@@ -91,6 +108,33 @@ class CommandLineHandler
...
+				processSensorsBehaviour = ProcessSensorsBehaviour::PSB_RENDER_ALL_THEN_INTERACTIVE; // default, when no behaviour option is specified
+
+				const auto& values = rawVariables[REA_PROCESS_SENSORS].value();
+				if (!values.empty())
+				{
+					const char* behaviour = values[0].c_str();
+					if (strcmp(behaviour, "RenderAllThenTerminate") == 0)
+					{
+						processSensorsBehaviour = ProcessSensorsBehaviour::PSB_RENDER_ALL_THEN_TERMINATE;
+						assert(values.size() == 1 && "Passing a sensor ID doens't make sense here.");
```
devshgraphicsprogramming, line 122: actually log an error and set a sane defaul, asserts (crashes) are not acceptable in this application

22.RaytracedAO/CommandLineHandler.hpp
```diff
@@ -91,6 +108,33 @@ class CommandLineHandler
...
+						processSensorsBehaviour = ProcessSensorsBehaviour::PSB_RENDER_SENSOR_THEN_INTERACTIVE;
+						sensorID = std::stoi(values[1]);
+					}
+					else if (strcmp(behaviour, "InteractiveAtSensor") == 0)
+					{
+						processSensorsBehaviour = ProcessSensorsBehaviour::PSB_INTERACTIVE_AT_SENSOR;
+						sensorID = std::stoi(values[1]);
+					}
+					else
+						assert(!"Invalid option.");
```
devshgraphicsprogramming, line 135: actually log an error and set a sane defaul, asserts (crashes) are not acceptable in this application

devshgraphicsprogramming:
remember to update README.md of example 22 as you're going

22.RaytracedAO/CommandLineHandler.hpp
achalpandeyy: It was just a sanity check. I don't think its an error because we can just ignore the additional passed value (perhaps with a warning).

22.RaytracedAO/CommandLineHandler.hpp
achalpandeyy: Resolved.

22.RaytracedAO/CommandLineHandler.hpp
```diff
@@ -16,69 +16,77 @@ constexpr std::string_view helpMessage = R"(
...
 	
 Example Usages :
-	raytracedao.exe -SCENE=../../media/kitchen.zip scene.xml -TERMINATE
-	raytracedao.exe -SCENE="../../media/my good kitchen.zip" scene.xml -TERMINATE
-	raytracedao.exe -SCENE="../../media/my good kitchen.zip scene.xml" -TERMINATE
-	raytracedao.exe -SCENE="../../media/extraced folder/scene.xml" -TERMINATE
+	raytracedao.exe -SCENE=../../media/kitchen.zip scene.xml
+	raytracedao.exe -SCENE="../../media/my good kitchen.zip" scene.xml
+	raytracedao.exe -SCENE="../../media/my good kitchen.zip scene.xml"
+	raytracedao.exe -SCENE="../../media/extraced folder/scene.xml"
```
devshgraphicsprogramming, line 29: keep the help-text and documentation up to date

achalpandeyy: Resolved.

22.RaytracedAO/CommandLineHandler.hpp
```diff
@@ -21,11 +21,16 @@ Description and usage:
...
 	some/path extra/path which will make it skip the file choose dialog
+
+-PROCESS_SENSORS ID:
+	It will control the behaviour of sensors in the app as detailed above.
+	If the option is not passed, then it defaults to RenderAllThenInteractive.
+	If the ID is not passed, then it defaults to 0.
 	
 Example Usages :
 	raytracedao.exe -SCENE=../../media/kitchen.zip scene.xml
 	raytracedao.exe -SCENE=../../media/kitchen.zip scene.xml -PROCESS_SENSORS RenderAllThenInteractive
```
devshgraphicsprogramming, line 32: you want `RenderAllThenInteractive` to also take an ID

22.RaytracedAO/main.cpp
```diff
@@ -754,7 +741,7 @@ int main(int argc, char** argv)
 	for(uint32_t s = 0u; s < globalMeta->m_global.m_sensors.size(); ++s)
 	{
 		if (s == sensorIndex)
-			continue; // skip it because we have already addes this one above
+			continue; // skip it because we have already added this one above
```
devshgraphicsprogramming, line 744: don't do this maybe, and add them all

then fix up the iteration at line 777/764 to start at `sensorIndex`

22.RaytracedAO/main.cpp
```diff
@@ -717,8 +725,24 @@ int main(int argc, char** argv)
...
+	uint32_t sensorIndex = cmdHandler.getSensorID();
+	if (sensorIndex >= globalMeta->m_global.m_sensors.size())
+	{
+		printf("[WARNING]: A valid sensor ID was not found. Selecting the first sensor.\n");
+		sensorIndex = 0;
+	}
+
+	// Add the found sensor to sensor data, if no (or invalid) ID was specified then just add the first sensor.
+	const auto& sensor = globalMeta->m_global.m_sensors[sensorIndex];
+	extractAndAddToSensorData(sensor, sensorIndex);
```
devshgraphicsprogramming, line 738: see the comment about adjusting iteration instead.

22.RaytracedAO/main.cpp
achalpandeyy: Yeah this would've been a great prediction, because I'm fixing a bug related to this.

The correct solution would be to add all and then save a "slice" into the sensor array for the non-interactive mode sensors.

22.RaytracedAO/CommandLineHandler.hpp
achalpandeyy: It can, but it cannot also (in which case it defaults 0). This usage shows when `RenderAllThenInteractive` doesn't take an ID.

22.RaytracedAO/main.cpp
achalpandeyy: Its already resolved, btw.

PR #46, opened 2023-06-01
Fix examples 27 & 42

deprilula28:
Fixes the examples with new API usages


27.PLYSTLDemo/main.cpp
```diff
@@ -24,7 +24,7 @@ using namespace core;
 	Uncomment for writing assets
 */
 
-#define WRITE_ASSETS
+//#define WRITE_ASSETS
```
devshgraphicsprogramming, line 27: how does that fix anything?

PR #49, opened 2023-07-13
Cuda interop

devshgraphicsprogramming:


63.CUDAInterop/main.cpp
```diff
@@ -0,0 +1,189 @@
...
+		auto buf2 = createBuffer(cpubuffers[2]);
+		
+		video::CCUDADevice::SSharedCUDAMemory mem0, mem1, mem2;
+
+		ASSERT_SUCCESS(cudaDevice->importGPUBuffer(buf0.get(), &mem0));
+		ASSERT_SUCCESS(cudaDevice->importGPUBuffer(buf1.get(), &mem1));
+		ASSERT_SUCCESS(cudaDevice->importGPUBuffer(buf2.get(), &mem2));
+
+		void* parameters[] = { &mem0.ptr, &mem1.ptr, &mem2.ptr, &numElements };
+		ASSERT_SUCCESS(cu.pcuLaunchKernel(kernel, gridDim[0], gridDim[1], gridDim[2], blockDim[0], blockDim[1], blockDim[2], 0, stream, parameters, nullptr));
```
devshgraphicsprogramming, line 84: I have a question, how do you deal with ownership transfers?

Cause by the time `createFilledDeviceLocalBufferOnDedMem` returns, the Vulkan queue which you used for transferring the data will own each buffer.

AFAIK, you need to do a pipeline barrier to an FOREIGN or EXTERNAL (does CUDA for the same device UUID count as external though?) queue before you start using it in CUDA.

63.CUDAInterop/main.cpp
```diff
@@ -0,0 +1,189 @@
...
+		ASSERT_SUCCESS(cu.pcuMemcpyDtoHAsync_v2(cpubuffers[2]->getPointer(), mem2.ptr, size, stream));
+		ASSERT_SUCCESS(cu.pcuCtxSynchronize());
```
devshgraphicsprogramming, line 86: IIRC this blocks the CPU, I'd eventually expect to be able to synchronize between CUDA and Nabla using an imported or exported semaphroe or some other synchronization primitive

63.CUDAInterop/main.cpp
```diff
@@ -0,0 +1,189 @@
...
+	{
+		video::CCUDADevice::SSharedCUDAMemory mem0, mem1, mem2;
+		ASSERT_SUCCESS(cudaDevice->createExportableMemory(size, sizeof(float), &mem0));
+		ASSERT_SUCCESS(cudaDevice->createExportableMemory(size, sizeof(float), &mem1));
+		ASSERT_SUCCESS(cudaDevice->createExportableMemory(size, sizeof(float), &mem2));
+
+		void* parameters[] = { &mem0.ptr, &mem1.ptr, &mem2.ptr, &numElements };
+		ASSERT_SUCCESS(cu.pcuMemcpyHtoDAsync_v2(mem0.ptr, cpubuffers[0]->getPointer(), size, stream));
+		ASSERT_SUCCESS(cu.pcuMemcpyHtoDAsync_v2(mem1.ptr, cpubuffers[1]->getPointer(), size, stream));
+		ASSERT_SUCCESS(cu.pcuCtxSynchronize());
```
devshgraphicsprogramming, line 126: why synchronize before launching a kernel?

devshgraphicsprogramming:
Very nice concise example

63.CUDAInterop/main.cpp
atkurtul: I thought it was needed in host to device copies. I guess we don't need to sync as long as cpu buffers stay in scope.

63.CUDAInterop/main.cpp
atkurtul: Noted

63.CUDAInterop/main.cpp
atkurtul: I'll need to figure out how to do transferships properly. Although validation layers does not complain, I'll poke around using different families to figure out.

`does CUDA for the same device UUID count as external though?`

I think it does.
From the spec:
The special queue family index VK_QUEUE_FAMILY_EXTERNAL represents any queue external to the resource’s current Vulkan instance, as long as the queue uses the same underlying device group or physical device, and the same driver version as the resource’s [VkDevice](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkDevice.html), as indicated by [VkPhysicalDeviceIDProperties](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceIDProperties.html)::deviceUUID and [VkPhysicalDeviceIDProperties](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceIDProperties.html)::driverUUID.

devshgraphicsprogramming:
superseded

PR #58, opened 2023-09-23
Colorspace test

Przemog1:


64.CppCompat/test.hlsl
```diff
@@ -0,0 +1,66 @@
...
+    nbl::hlsl::colorspace::oetf::sRGB<float3>(TEST_VEC);
+    nbl::hlsl::colorspace::oetf::Display_P3<float3>(TEST_VEC);
+    nbl::hlsl::colorspace::oetf::DCI_P3_XYZ<float3>(TEST_VEC);
+    nbl::hlsl::colorspace::oetf::SMPTE_170M<float3>(TEST_VEC);
+    nbl::hlsl::colorspace::oetf::SMPTE_ST2084<float3>(TEST_VEC);
+    nbl::hlsl::colorspace::oetf::HDR10_HLG<float3>(TEST_VEC);
+    nbl::hlsl::colorspace::oetf::AdobeRGB<float3>(TEST_VEC);
+    nbl::hlsl::colorspace::oetf::Gamma_2_2<float3>(TEST_VEC);
+    nbl::hlsl::colorspace::oetf::ACEScc<float3>(TEST_VEC);
+    nbl::hlsl::colorspace::oetf::ACEScct<float3>(TEST_VEC);
```
Przemog1, line 56: how do i test if these output correct value?

devshgraphicsprogramming: generally, 0.f should map to 0.f and for most OETF/EOTFs (there are some exceptions) 1.f to 1.f

IIRC the only "exception" is scRGB which allows "out of gamut values" but everything else must stay in  a [0,1]^3 range on input and output

I guess you can just tabulate some "special" values, i.e. what these functions spit out at 0.f, 0.2f, ... , 1.f

64.CppCompat/main.cpp
```diff
@@ -11,13 +11,149 @@
...
+    glm::vec3(1.0498110175f, -0.4959030231f, 0.0000000000f),
+    glm::vec3(0.0000000000f, 1.3733130458f, 0.0000000000f),
+    glm::vec3(-0.0000974845f, 0.0982400361f, 0.9912520182f)
+);
+
+constexpr glm::mat3 nbl_glsl_XYZtoACEScc = glm::mat3(
+    glm::vec3(1.6410233797f, -0.6636628587f, 0.0117218943f),
+    glm::vec3(-0.3248032942f, 1.6153315917f, -0.0082844420f),
+    glm::vec3(-0.2364246952f, 0.0167563477f, 0.9883948585f)
+);
```
devshgraphicsprogramming, line 113: why not just `#include` the `builtin/glsl` files?

64.CppCompat/main.cpp
```diff
@@ -36,10 +172,132 @@ struct T
...
+    std::array<video::IGPUQueue*, CommonAPI::InitOutput::MaxQueuesCount> m_queues;
+    core::smart_refctd_ptr<video::IGPUSemaphore> m_semaphores[2u] = {nullptr};
+    core::smart_refctd_ptr<video::IGPUFence> m_fence = nullptr;
+    std::array<std::array<nbl::core::smart_refctd_ptr<nbl::video::IGPUCommandPool>, CommonAPI::InitOutput::MaxFramesInFlight>, CommonAPI::InitOutput::MaxQueuesCount> commandPools;
+    core::smart_refctd_ptr<nbl::asset::IAssetManager> m_assetManager;
+    video::IGPUObjectFromAssetConverter::SParams m_cpu2gpuParams;
+    core::smart_refctd_ptr<nbl::system::ILogger> m_logger;
+
+    bool m_keepRunning = true;
+};
```
devshgraphicsprogramming, line 291: noooo, this needs to be a 100% GPU-less test

64.CppCompat/test.hlsl
devshgraphicsprogramming: btw don't test in a compute shader, do all tests in c++

64.CppCompat/main.cpp
Przemog1: then how do we check if it works with hlsl?

devshgraphicsprogramming: > then how do we check if it works with hlsl?

our cpp_compat should already give you 1:1 with HLSL when math is concerned

Przemog1: it is usefull for me so will leave it for now

devshgraphicsprogramming:
you need to merge the tests from `master` because of conflicts

PR #70, opened 2023-11-02
Fix sorting comparator problem & AA

deprilula28:


62.CAD/Hatch.cpp
```diff
@@ -1,52 +1,233 @@
...
-	// If either is NaN or both are equal
-	// Same as: 
-	// if (t[0] == t[1] || isnan(t[0]) || isnan(t[1]))
-	if (!(t[0] != t[1]))
-		t[0] = t[0] != t_start ? t_start : t_end;
+	return roots;
+}
+
+// Intended to mitigate issues with NaNs and precision by falling back to using simpler functions when the higher roots are small enough
+std::array<double, 4> Hatch::solveQuarticRoots(double a, double b, double c, double d, double e, double t_start, double t_end)
```
Erfan-Ahmadi, line 169: you don't need the `double t_start, double t_end` anymore?

PR #77, opened 2023-11-28
Cad example work

Erfan-Ahmadi:


common/CommonAPI.h
```diff
@@ -1,20 +1,29 @@
 #ifndef __NBL_COMMON_API_H_INCLUDED__
 #define __NBL_COMMON_API_H_INCLUDED__
 
+#define _NBL_STATIC_LIB_
```
devshgraphicsprogramming, line 4: can you live without this?

Erfan-Ahmadi: don't remember what that is, but when I remove it, it still works :D

devshgraphicsprogramming: remove it, its a define that should only be defined when CMake is in the static config, shoul dbe already in `BuildConfig.h` which is autogeneretated

AnastaZIuk: I believe a lot of examples were defining it, may be a remnant of very old code. `_NBL_SHARED_BUILD_` is defined instead https://github.com/Devsh-Graphics-Programming/Nabla/blob/master/include/nbl/config/BuildConfigOptions.h.in#L90

PR #80, opened 2023-12-06
CPP Compat test renumber and refactor

devshgraphicsprogramming:


CMakeLists.txt
```diff
@@ -36,7 +36,7 @@ if(NBL_BUILD_EXAMPLES)
 	# Unit Test Examples
 	add_subdirectory(20_AllocatorTest EXCLUDE_FROM_ALL)
 	add_subdirectory(21_LRUCacheUnitTest EXCLUDE_FROM_ALL)
-
+    add_subdirectory(22_CppCompat EXCLUDE_FROM_ALL)
```
devshgraphicsprogramming, line 39: remove the commented out line with 64 on line 69

common/MonoAssetManagerAndBuiltinResourceApplication.hpp
```diff
@@ -10,6 +10,8 @@
 #include "nbl/this_example/builtin/CArchive.h"
 #endif
 
+#include "nbl\system\CMountDirectoryArchive.h"
+
```
devshgraphicsprogramming, line 14: this should come in from `nabla.h` more specifically `nbl/system.h` (or def/decl if it has them)

PR #81, opened 2023-12-14
Refactor the FilterTest to use the new common examples API

theoreticalphysicsftw:
Use `MonoDeviceApplication` as base for the example.


25_FilterTest/main.cpp
```diff
@@ -1585,78 +1608,23 @@ class BlitFilterTestApp : public ApplicationBase
...
+		retval.vulkanMemoryModelDeviceScope = true;
+
+		return retval;
 	}
 
-	APP_CONSTRUCTOR(BlitFilterTestApp);
+	core::smart_refctd_ptr<asset::IAssetManager> assetManager;
+	video::IGPUQueue* queue;
+	core::smart_refctd_ptr<video::IGPUCommandPool> commandPool;
+	core::smart_refctd_ptr<video::IUtilities> utilities;
```
devshgraphicsprogramming, line 1623: maybe inherit from `BasicMultiQueueApplication` instead of `MonoDeviceApplication` because you use `IUtilities`

due to asset manager usage I also recommend `MonoAssetManagerAndBuiltinResourceApplication` inheritance

theoreticalphysicsftw: Actually trying to use `BasicMultiQueueApplication` I realized that the way `IGPUObjectFromAssetConverter` works now it requires both the compute and transfer queue to also support graphics, while `BasicMultiQueueApplication` tries to pick-up the Async-Compute and DMA engines first.

devshgraphicsprogramming: > Actually trying to use `BasicMultiQueueApplication` I realized that the way `IGPUObjectFromAssetConverter` works now it requires both the compute and transfer queue to also support graphics, while `BasicMultiQueueApplication` tries to pick-up the Async-Compute and DMA engines first.

yes its a known bug, BECAUSE our asset converter is not using the compute queue to do mip-map generation :)

Next thing on your TODO will be to make it use the HLSL Blit instead, then there will no longer be a requirement for a graphics capable queue from `IGPUObjectFromAssetConverter`

P.S. IIRC its only the compute queue that needs to support graphics.

PR #82, opened 2023-12-15
Fix lack of graphics queue in non-headless-compute `BasicMultiQueueApplication`

theoreticalphysicsftw:
Right now when `isHeadlessCompute` is overriden to return false the `m_graphicsQueue` is still uninitialized which leads `getGraphicsQueue()` to fail. This PR inits the graphics queue when `!isHeadlessCompute() == true`.

### DIFF
--- a/common/BasicMultiQueueApplication.hpp
+++ b/common/BasicMultiQueueApplication.hpp
@@ -173,6 +173,14 @@ class BasicMultiQueueApplication : public virtual MonoDeviceApplication
 			else
 				m_transferDownQueue.qIx = familyQueueCounts[m_transferDownQueue.famIx]++;
 
+			if (!isHeadlessCompute())
+			{
+				queue_req_t graphicsQueueRequirement = { .requiredFlags = queue_flags_t::EQF_GRAPHICS_BIT,.disallowedFlags = queue_flags_t::EQF_NONE,.queueCount = 1 };
+				m_graphicsQueue.famIx = queueAllocator.allocateFamily(graphicsQueueRequirement, { queue_flags_t::EQF_NONE });
+				assert(m_graphicsQueue.famIx != QueueAllocator::InvalidIndex);
+				m_graphicsQueue.qIx = familyQueueCounts[m_graphicsQueue.famIx]++;
+			}
+
 			// now after assigning all queues to families and indices, collate the creation parameters
 			core::vector<video::ILogicalDevice::SQueueCreationParams> retval(familyQueueCounts.size());
 			auto oit = retval.begin();
--- a/common/BasicMultiQueueApplication.hpp
+++ b/common/BasicMultiQueueApplication.hpp
@@ -121,11 +121,30 @@ class BasicMultiQueueApplication : public virtual MonoDeviceApplication
 		{
 			QueueAllocator queueAllocator(familyProperties);
 
-			// First thing to make sure we have is a compute queue (so nothing else fails allocation) which should be able to do image transfers of any granularity (transfer only queue families can have problems with that)
-			queue_req_t computeQueueRequirement = {.requiredFlags=queue_flags_t::EQF_COMPUTE_BIT,.disallowedFlags=queue_flags_t::EQF_NONE,.queueCount=1,.maxImageTransferGranularity={1,1,1}};
-			m_computeQueue.famIx = queueAllocator.allocateFamily(computeQueueRequirement,{queue_flags_t::EQF_GRAPHICS_BIT,queue_flags_t::EQF_TRANSFER_BIT,queue_flags_t::EQF_SPARSE_BINDING_BIT,queue_flags_t::EQF_PROTECTED_BIT});
-			// since we requested a device that has a compute capable queue family (unless `getQueueRequirements` got overriden) we're sure we'll get at least one family capable of compute
-			assert(m_computeQueue.famIx!=QueueAllocator::InvalidIndex);
+			// This is a sort-of allocator of queue indices for distinct queues
+			core::map<uint8_t, uint8_t> familyQueueCounts;
+
+			if (!isHeadlessCompute())
+			{
+				// TODO: Handle swapchain compatibility.
+				queue_req_t graphicsQueueRequirement = { .requiredFlags = queue_flags_t::EQF_GRAPHICS_BIT,.disallowedFlags = queue_flags_t::EQF_NONE,.queueCount = 1 };
+				// Place queue_flags_t::EQF_NONE in front in case there are no graphics queues without the unwanted flags.
+				m_graphicsQueue.famIx = queueAllocator.allocateFamily(graphicsQueueRequirement, { queue_flags_t::EQF_NONE, queue_flags_t::EQF_TRANSFER_BIT,queue_flags_t::EQF_SPARSE_BINDING_BIT, queue_flags_t::EQF_COMPUTE_BIT, queue_flags_t::EQF_PROTECTED_BIT });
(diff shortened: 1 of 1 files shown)


common/BasicMultiQueueApplication.hpp
```diff
@@ -173,6 +173,14 @@ class BasicMultiQueueApplication : public virtual MonoDeviceApplication
...
 				m_transferDownQueue.qIx = familyQueueCounts[m_transferDownQueue.famIx]++;
 
+			if (!isHeadlessCompute())
+			{
+				queue_req_t graphicsQueueRequirement = { .requiredFlags = queue_flags_t::EQF_GRAPHICS_BIT,.disallowedFlags = queue_flags_t::EQF_NONE,.queueCount = 1 };
+				m_graphicsQueue.famIx = queueAllocator.allocateFamily(graphicsQueueRequirement, { queue_flags_t::EQF_NONE });
+				assert(m_graphicsQueue.famIx != QueueAllocator::InvalidIndex);
+				m_graphicsQueue.qIx = familyQueueCounts[m_graphicsQueue.famIx]++;
+			}
+
```
devshgraphicsprogramming, line 183: this needs a better fix, you're right that I forgot to allocate the graphics queue.

The reason was because Graphics Queue Families need to be checked for compatibility with a swapchain.

For later examples I need to either rework the `allocateFamily` function to take a template lambda instead of `originalReq`, OR extend `SPhysicalDeviceFilter::QueueRequirement` to also include a list of surfaces the family must be able to present to.

devshgraphicsprogramming: ok so about the better fix

If you notice, I have to handle allocation failures gracefully.

This is because old (non ARC) Intels only have one queue family that can provide only a single queue, therefore I have to alias failures.

So here I'd allocate the Graphics queue first, and assert it must have successfully allocated. Then if you fail to allocate a different queue for Compute, you should alias the compute queue to the graphics one.

Read the existing code see how it falls back and logs everything, follow similar convention.

You should also provide a list of unwanted flags for the Graphics queue allocation in order of most unwanted to least, so TRANSFER, SPARSE_BIND, COMPUTE, PROTECTED,

common/BasicMultiQueueApplication.hpp
```diff
@@ -121,11 +121,30 @@ class BasicMultiQueueApplication : public virtual MonoDeviceApplication
...
-			assert(m_computeQueue.famIx!=QueueAllocator::InvalidIndex);
+			// This is a sort-of allocator of queue indices for distinct queues
+			core::map<uint8_t, uint8_t> familyQueueCounts;
+
+			if (!isHeadlessCompute())
+			{
+				// TODO: Handle swapchain compatibility.
+				queue_req_t graphicsQueueRequirement = { .requiredFlags = queue_flags_t::EQF_GRAPHICS_BIT,.disallowedFlags = queue_flags_t::EQF_NONE,.queueCount = 1 };
+				// Place queue_flags_t::EQF_NONE in front in case there are no graphics queues without the unwanted flags.
+				m_graphicsQueue.famIx = queueAllocator.allocateFamily(graphicsQueueRequirement, { queue_flags_t::EQF_NONE, queue_flags_t::EQF_TRANSFER_BIT,queue_flags_t::EQF_SPARSE_BINDING_BIT, queue_flags_t::EQF_COMPUTE_BIT, queue_flags_t::EQF_PROTECTED_BIT });
```
devshgraphicsprogramming, line 132: EQF NONE is just a 0, it won't do anything

common/BasicMultiQueueApplication.hpp
```diff
@@ -121,11 +121,30 @@ class BasicMultiQueueApplication : public virtual MonoDeviceApplication
...
+			queue_req_t computeQueueRequirement = { .requiredFlags = queue_flags_t::EQF_COMPUTE_BIT,.disallowedFlags = queue_flags_t::EQF_NONE,.queueCount = 1,.maxImageTransferGranularity = {1,1,1} };
+			m_computeQueue.famIx = queueAllocator.allocateFamily(computeQueueRequirement, { queue_flags_t::EQF_GRAPHICS_BIT,queue_flags_t::EQF_TRANSFER_BIT,queue_flags_t::EQF_SPARSE_BINDING_BIT,queue_flags_t::EQF_PROTECTED_BIT });
+
+			if (m_computeQueue.famIx == QueueAllocator::InvalidIndex)
+			{
+				// Since we requested a device that has a compute capable queue family (unless `getQueueRequirements` got overriden) we're sure we'll get at least one family capable of compute.
+				// Going through this branch means we've selected that family for the graphics queue and no more queues can be allocated for this family. We have to alias.
+				m_logger->log("Not enough queue counts in families, had to alias the Compute Queue to Graphics!", system::ILogger::ELL_PERFORMANCE);
+				m_computeQueue = m_graphicsQueue;
+			}
```
devshgraphicsprogramming, line 147: you need an else clause on both  the if-statements, because for non-headless you want to disallow the GRAPHICS queue capability when allocating the compute queue

devshgraphicsprogramming: Actually best thing would be to allocate a fallback queue first that has following caps
```cpp
queue_req_t{ .requiredFlags = (isHeadlessCompute() ? EQF_GRAPHICS:EQF_NONE)|queue_flags_t::EQF_COMPUTE_BIT|queue_flags_t::EQF_transFER_BIT,.disallowedFlags = queue_flags_t::EQF_NONE,.queueCount = 1,.maxImageTransferGranularity = {1,1,1} };
```

then try-allocate separate transfer queues (with fallback of download to upload first), then try-allocate a separate compute queue without graphics capabilities, and finally a graphics queue without all the extra FORCED constraints

common/BasicMultiQueueApplication.hpp
theoreticalphysicsftw: No it's actually important, being 0 is exactly the point. The way `allocateFamily` works it tries to find a queue without all the flags, then removes one flag and tries again, until it reaches the state with just 1 unwanted flag (the first one). And if that flag is still present in all queues it will fail. That's what's actually happening on my RX 6600 (because there is no queue without the transfer flag). So the last flag `EQF_NONE` is for `allocateFamily` to do one last iteration where it tries to find a graphics queue without any restrictions.

common/BasicMultiQueueApplication.hpp
theoreticalphysicsftw: > because for non-headless you want to disallow the GRAPHICS queue capability when allocating the compute queue, and you no longer need to require transfer granularities of 1,1,1

Yeah you're right about the transfer granularities but the current code already tries to find a compute queue without graphics in non-headless and only fallback to graphics capable queue if no such compute queue exists (for example having only combined graphics + compute and no async one). Or in such case the idea is (if no compute only queue exists) that we should just alias and not allocate another queue that also has graphics capabilities?

theoreticalphysicsftw: >Actually best thing would be to allocate a fallback queue first that has following caps

Yeah that would simplify things. Pick one good for everything queue and then if we don't find async compute, DMA and graphics only queue for each alias to the fallback.

common/BasicMultiQueueApplication.hpp
devshgraphicsprogramming: ok I see what you mean, the loop will not execute
```cpp
					for (size_t flagCount=unwantedFlags.size(); flagCount; flagCount--)
					{
						queue_req_t req = originalReq;
						for (auto it=unwantedFlags.begin(); it!=(unwantedFlags.begin()+flagCount); it++)
```

I'm rewriting to a `do{}while();`

PR #84, opened 2023-12-18
Example 06: Hello graphics queue

Przemog1:


06_HelloGraphicsQueue/main.cpp
```diff
@@ -2,80 +2,436 @@
...
-		//
-		bool keepRunning() override {return false;}
+	IGPUCommandBuffer::SImageMemoryBarrier constructDefaultInitializedImageBarrier() const
+	{
+		IImage::SSubresourceRange imgSubresourceRange{};
+		imgSubresourceRange.aspectMask = IImage::E_ASPECT_FLAGS::EAF_COLOR_BIT;
+		imgSubresourceRange.baseMipLevel = 0u;
+		imgSubresourceRange.levelCount = 1u;
+		imgSubresourceRange.baseArrayLayer = 0u;
+		imgSubresourceRange.layerCount = 1u;
```
Przemog1, line 357: remove

06_HelloGraphicsQueue/main.cpp
```diff
@@ -2,80 +2,436 @@
...
+		if (!device_base_t::onAppInitialized(std::move(system)))
+			return false;
+		if (!asset_base_t::onAppInitialized(std::move(system)))
+			return false;
 
-	public:
-		// Yay thanks to multiple inheritance we cannot forward ctors anymore
-		HelloGraphicsQueueApp(const path& _localInputCWD, const path& _localOutputCWD, const path& _sharedInputCWD, const path& _sharedOutputCWD) :
-			system::IApplicationFramework(_localInputCWD,_localOutputCWD,_sharedInputCWD,_sharedOutputCWD) {}
+		IGPUQueue* const queue = getQueue(IPhysicalDevice::E_QUEUE_FLAGS::EQF_GRAPHICS_BIT | IPhysicalDevice::E_QUEUE_FLAGS::EQF_TRANSFER_BIT);
```
devshgraphicsprogramming, line 45: write a comment about why you're getting such a queue

06_HelloGraphicsQueue/main.cpp
```diff
@@ -2,80 +2,436 @@
...
+			outputBufferAllocation = m_device->allocate(reqs, outputImageBuffer.get(), nbl::video::IDeviceMemoryAllocation::E_MEMORY_ALLOCATE_FLAGS::EMAF_NONE);
+			if (!outputBufferAllocation.isValid())
+				return logFail("Failed to allocate Device Memory compatible with our GPU Buffer!\n");
+
+			// Note that we performed a Dedicated Allocation above, so there's no need to bind the memory anymore (since the allocator knows the dedication, it can already bind).
+			// This is a carryover from having an OpenGL backend, where you couldn't have a memory allocation separate from the resource, so all allocations had to be "dedicated".
+			// In Vulkan dedicated allocations are the most performant and still make sense as long as you won't blow the 4096 allocation limit on windows.
+			// You should always use dedicated allocations for images used for swapchains, framebuffer attachments (esp transient), as well as objects used in CUDA/DirectX interop.
+			assert(outputImageBuffer->getBoundMemory() == outputBufferAllocation.memory.get());
+		}
```
devshgraphicsprogramming, line 96: you don't need to copy comments from previous examples, only explain things that are new

06_HelloGraphicsQueue/main.cpp
```diff
@@ -2,80 +2,436 @@
...
+		IGPUCommandBuffer::SImageMemoryBarrier imgLayoutTransitionBarrier = constructDefaultInitializedImageBarrier();
+		imgLayoutTransitionBarrier.newLayout = asset::IImage::EL_TRANSFER_DST_OPTIMAL;
+		imgLayoutTransitionBarrier.barrier.dstAccessMask = E_ACCESS_FLAGS::EAF_MEMORY_WRITE_BIT;
+
+		IGPUCommandBuffer::SImageMemoryBarrier imgClearBarrier = constructDefaultInitializedImageBarrier();
+		imgClearBarrier.oldLayout = asset::IImage::EL_TRANSFER_DST_OPTIMAL;
+		imgClearBarrier.barrier.srcAccessMask = E_ACCESS_FLAGS::EAF_MEMORY_WRITE_BIT;
+
+		constexpr SClearColorValue red = { .float32{1.0f, 0.0f, 0.0f, 1.0f} };
+		constexpr SClearColorValue blue = { .float32{0.0f, 0.0f, 1.0f, 1.0f} };
```
devshgraphicsprogramming, line 158: explain why we're using this union member isntead of others (colorformat stuff I explained to you)

06_HelloGraphicsQueue/main.cpp
```diff
@@ -2,80 +2,436 @@
...
+			E_DEPENDENCY_FLAGS::EDF_NONE, 0u, nullptr, 0u, nullptr, 1u,
+			&bigImgLayoutTransitionBarrier
+		);
+
+		cmdbuf->pipelineBarrier(
+			E_PIPELINE_STAGE_FLAGS::EPSF_HOST_BIT | E_PIPELINE_STAGE_FLAGS::EPSF_ALL_COMMANDS_BIT,
+			E_PIPELINE_STAGE_FLAGS::EPSF_TRANSFER_BIT,
+			E_DEPENDENCY_FLAGS::EDF_NONE, 0u, nullptr, 0u, nullptr, 1u,
+			&smallImgLayoutTransitionBarrier
+		);
```
devshgraphicsprogramming, line 186: you should combine both barriers into one, after all  the command takes a _count_ of buffer and image barriers

06_HelloGraphicsQueue/main.cpp
```diff
@@ -2,80 +2,436 @@
...
+			E_PIPELINE_STAGE_FLAGS::EPSF_TRANSFER_BIT,
+			E_PIPELINE_STAGE_FLAGS::EPSF_TRANSFER_BIT,
+			E_DEPENDENCY_FLAGS::EDF_NONE, 0u, nullptr, 0u, nullptr, 1u, &bigImgClearBarrier
+		);
+
+		cmdbuf->pipelineBarrier(
+			E_PIPELINE_STAGE_FLAGS::EPSF_TRANSFER_BIT,
+			E_PIPELINE_STAGE_FLAGS::EPSF_TRANSFER_BIT,
+			E_DEPENDENCY_FLAGS::EDF_NONE, 0u, nullptr, 0u, nullptr, 1u, &smallImgClearBarrier
+		);
```
devshgraphicsprogramming, line 214: again, combine

06_HelloGraphicsQueue/main.cpp
```diff
@@ -2,80 +2,436 @@
...
+		bigImgBlitBarrier.oldLayout = asset::IImage::EL_TRANSFER_DST_OPTIMAL;
+		bigImgBlitBarrier.newLayout = asset::IImage::EL_TRANSFER_SRC_OPTIMAL;
+		bigImgBlitBarrier.image = bigImg;
+
+		IGPUCommandBuffer::SImageMemoryBarrier smallImgBlitBarrier = constructDefaultInitializedImageBarrier();
+		smallImgBlitBarrier.barrier.srcAccessMask = E_ACCESS_FLAGS::EAF_MEMORY_READ_BIT;
+		smallImgBlitBarrier.barrier.dstAccessMask = E_ACCESS_FLAGS::EAF_MEMORY_WRITE_BIT;
+		smallImgBlitBarrier.oldLayout = asset::IImage::EL_TRANSFER_SRC_OPTIMAL;
+		smallImgBlitBarrier.newLayout = asset::IImage::EL_TRANSFER_DST_OPTIMAL;
+		smallImgBlitBarrier.image = smallImg;
```
devshgraphicsprogramming, line 231: btw where's that lambda for joining two halves of pipeline barriers / generating the next one?

06_HelloGraphicsQueue/main.cpp
```diff
@@ -2,80 +2,436 @@
...
+		{
+			params.image = ICPUImage::create(ouputImageCreationParams);
+			{
+				auto cpuView = core::make_smart_refctd_ptr<CDummyCPUBuffer>(smallImgByteSize, imageBufferMemPtr, core::adopt_memory_t());
+				ICPUImage::SBufferCopy region = {};
+				region.imageSubresource.aspectMask = IImage::E_ASPECT_FLAGS::EAF_COLOR_BIT;
+				region.imageSubresource.layerCount = 1;
+				region.imageExtent = ouputImageCreationParams.extent;
+				params.image->setBufferAndRegions(std::move(cpuView),core::make_refctd_dynamic_array<core::smart_refctd_dynamic_array<ICPUImage::SBufferCopy>>(1,region));
+			}
```
devshgraphicsprogramming, line 297: this needs a lot of comments to explain whats going on

06_HelloGraphicsQueue/main.cpp
```diff
@@ -2,80 +2,436 @@
...
+			params.viewType = ICPUImageView::ET_2D;
+			params.format = ouputImageCreationParams.format;
+			params.subresourceRange.aspectMask = IImage::E_ASPECT_FLAGS::EAF_COLOR_BIT;
+			params.subresourceRange.layerCount = 1;
 		}
+		auto cpuImageView = ICPUImageView::create(std::move(params));
+		asset::IAssetWriter::SAssetWriteParams writeParams(cpuImageView.get());
+		m_assetMgr->writeAsset("blit.png", writeParams);
+
+		m_device->unmapMemory(outputBufferAllocation.memory.get());
```
devshgraphicsprogramming, line 308: you can write a comment that even if you forgot to unmap it, it would unmap itself when `outputBufferAllocation.memory` gets dropped by its last reference and its destructor runs

06_HelloGraphicsQueue/main.cpp
```diff
@@ -2,80 +2,436 @@
...
+	void workLoopBody() override {}
+
+	//
+	bool keepRunning() override {return false;}
+
+protected:
+	core::vector<queue_req_t> getQueueRequirements() const override
+	{
+		using flags_t = IPhysicalDevice::E_QUEUE_FLAGS;
+		return {{.requiredFlags=flags_t::EQF_GRAPHICS_BIT,.disallowedFlags=flags_t::EQF_NONE,.queueCount=1,.maxImageTransferGranularity={1,1,1}}};
```
devshgraphicsprogramming, line 323: if you want a GRAPHICS+TRANSFER queue with `getQueue` then you MUST declare this here

06_HelloGraphicsQueue/main.cpp
```diff
@@ -2,80 +2,436 @@
...
+			return nullptr;
 		}
+		auto allocation = m_device->allocate(imageBuffer->getMemoryReqs(), imageBuffer.get(), nbl::video::IDeviceMemoryAllocation::EMAF_NONE);
+
+		ILogicalDevice::SBindImageMemoryInfo bindMemInfo;
+		bindMemInfo.image = img.get();
+		bindMemInfo.memory = allocation.memory.get();
+		bindMemInfo.offset = 0ull;
+
+		m_device->bindImageMemory(1u, &bindMemInfo);
```
devshgraphicsprogramming, line 430: make sure you use dedicated allocations for the image

06_HelloGraphicsQueue/main.cpp
```diff
@@ -164,82 +164,49 @@ class HelloGraphicsQueueApp final : public examples::MonoDeviceApplication, publ
...
-			&bigImgLayoutTransitionBarrier
-		);
-
-		cmdbuf->pipelineBarrier(
-			E_PIPELINE_STAGE_FLAGS::EPSF_HOST_BIT | E_PIPELINE_STAGE_FLAGS::EPSF_ALL_COMMANDS_BIT,
-			E_PIPELINE_STAGE_FLAGS::EPSF_TRANSFER_BIT,
-			E_DEPENDENCY_FLAGS::EDF_NONE, 0u, nullptr, 0u, nullptr, 1u,
-			&smallImgLayoutTransitionBarrier
+			E_DEPENDENCY_FLAGS::EDF_NONE, 0u, nullptr, 0u, nullptr, 2u,
+			imgLayoutTransitionBarriers.data()
```
devshgraphicsprogramming, line 179: its good practice to write `.size()` instead of a hardcoded literal, esp if the container knows its size anyway

06_HelloGraphicsQueue/main.cpp
```diff
@@ -150,10 +139,21 @@ class HelloGraphicsQueueApp final : public examples::MonoDeviceApplication, publ
...
+				IGPUCommandBuffer::SImageMemoryBarrier ret = prevoiusBarrier;
+				ret.oldLayout = prevoiusBarrier.newLayout;
+				ret.newLayout = newLayout;
+				ret.barrier.srcAccessMask = prevoiusBarrier.barrier.dstAccessMask;
+				ret.barrier.dstAccessMask = dstAccessMast;
+
+				return ret;
+			};
 
+		// Using `float32` union member since format of used images is E_FORMAT::EF_R8G8B8A8_SRGB.
```
devshgraphicsprogramming, line 156: add "SRGB, which is a fixed-point format and those are encoded/decoded from/to float"

06_HelloGraphicsQueue/main.cpp
devshgraphicsprogramming: apply this to allthe sites where you do this

06_HelloGraphicsQueue/main.cpp
devshgraphicsprogramming: basically explain that our ICPUImage isn't really a representation of a GPU Image in itself, more of a recipe for creating one from a series of ICPUBuffer to ICPUImage copies.

This means that an ICPUImage has no internal storage or memory bound for its texels and rather references separate ICPUBuffer ranges to provide its contents, which also means it can be sparsely (with gaps) specified.

06_HelloGraphicsQueue/main.cpp
devshgraphicsprogramming: In general on Windows, you get a hard limit of 4096 vkDeviceMemory allocations.

Which is why you're supposed to combine and suballocate multiple resources from same memory as long as your heap requirements match (type and attributes).

However, because:
- some non-Vulkan APIs don't have a concept of separable resource and memory allocation binding (i.e. when we used GL in the past) or
- you're importing/exporting images and buffers to/from DX12, CUDA, etc., or
- there's a performance improvement (i.e. Nvidia framebuffer images, or self-created swapchain images)
you might have to use dedicated allocations (only 1 resource binds to a particular vkDevicememory)

In Nabla we do it by passing the resource the allocation will be dedicated to as part of the memory allocation method:
https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/blob/aff738da77804a39684c3053867d77b03c5fd803/02_HelloCompute/main.cpp#L195

Here you requested a dedication, which means the memory is ALREADY bound, the latter `m_device->bindImageMemory` won't actually succeed and you'll get an error, its not needed.

PR #94, opened 2024-01-25
Property pool hlsl

deprilula28:
Sister PR : https://github.com/Devsh-Graphics-Programming/Nabla/pull/649

PR #98, opened 2024-03-04
Example 26 box blur

Eichenherz:


26_CentralLimitBoxBlur/main.cpp
```diff
@@ -0,0 +1,339 @@
...
+	{}
+	
+	bool onAppInitialized( smart_refctd_ptr<ISystem>&& system ) override
+	{
+		// Remember to call the base class initialization!
+		if( !device_base_t::onAppInitialized( std::move( system ) ) )
+		{
+			return false;
+		}
+		if( !asset_base_t::onAppInitialized( std::move( system ) ) )
```
devshgraphicsprogramming, line 43: you can't move the same smart pointer twice, you need to copy it, see a different example

common/MonoDeviceApplication.hpp
```diff
@@ -18,7 +18,7 @@ class MonoDeviceApplication : public virtual MonoSystemMonoLoggerApplication
 	public:
 		using base_t::base_t;
 
-	protected:
+	public:
```
devshgraphicsprogramming, line 21: change not needed, you've inheriting from it always.

CMakeLists.txt
```diff
@@ -65,5 +65,6 @@ if(NBL_BUILD_EXAMPLES)
 	#add_subdirectory(61_UI EXCLUDE_FROM_ALL)
 	add_subdirectory(62_CAD EXCLUDE_FROM_ALL)
 	add_subdirectory(62_SchusslerTest EXCLUDE_FROM_ALL)
+	add_subdirectory(66_PropertyPools EXCLUDE_FROM_ALL)
```
devshgraphicsprogramming, line 68: why do I have parts of #94 in this PR?

PR #87, opened 2023-12-23
Implemened example 07

Przemog1:

PR #99, opened 2024-03-04
Vulkan_1_3 example 07 fix

Przemog1:
Old PR #87


07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
 // This file is part of the "Nabla Engine".
 // For conditions of distribution and use, see copyright notice in nabla.h
 
-
 // I've moved out a tiny part of this example into a shared header for reuse, please open and read it.
 #include "../common/BasicMultiQueueApplication.hpp"
 #include "../common/MonoAssetManagerAndBuiltinResourceApplication.hpp"
 
+//std::this_thread::sleep_for(std::chrono::seconds(1)); // TODO: remove
```
devshgraphicsprogramming, line 9: huh?

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
 #include "app_resources/common.hlsl"
 
-
 // This time we let the new base class score and pick queue families, as well as initialize `nbl::video::IUtilities` for us
 class StagingAndMultipleQueuesApp final : public examples::BasicMultiQueueApplication, public examples::MonoAssetManagerAndBuiltinResourceApplication
 {
-		using device_base_t = examples::BasicMultiQueueApplication;
-		using asset_base_t = examples::MonoAssetManagerAndBuiltinResourceApplication;
+	using device_base_t = examples::BasicMultiQueueApplication;
+	using asset_base_t = examples::MonoAssetManagerAndBuiltinResourceApplication;
```
devshgraphicsprogramming, line 23: merge latest master and adjust, the framework applications changed namespace

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+	StagingAndMultipleQueuesApp(const path& _localInputCWD, const path& _localOutputCWD, const path& _sharedInputCWD, const path& _sharedOutputCWD) :
+		system::IApplicationFramework(_localInputCWD, _localOutputCWD, _sharedInputCWD, _sharedOutputCWD) {}
+
+	// This time we will load images and compute their histograms and output them as CSV
+	bool onAppInitialized(smart_refctd_ptr<ISystem>&& system) override
+	{
+		// Remember to call the base class initialization!
+		if (!device_base_t::onAppInitialized(std::move(system)))
+			return false;
+		if (!asset_base_t::onAppInitialized(std::move(system)))
```
devshgraphicsprogramming, line 49: a smart pointer can only be moved once, you need to copy the first time round

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+		return retval;
+	}
+
+	// Ideally don't want to have to 
+	SPhysicalDeviceFeatures getPreferredDeviceFeatures() const override
+	{
+		auto retval = device_base_t::getPreferredDeviceFeatures();
+		//retval.shaderStorageImageReadWithoutFormat = true;
+		return retval;
+	}
```
devshgraphicsprogramming, line 99: you can just delete the overrides

07_StagingAndMultipleQueues/app_resources/comp_shader.hlsl
```diff
@@ -0,0 +1,34 @@
+#pragma wave shader_stage(compute)
+
+#include "../app_resources/common.hlsl"
+
+static const uint32_t RED_OFFSET = 0u;
+static const uint32_t GREEN_OFFSET = 256u;
+static const uint32_t BLUE_OFFSET = 256u * 2u;
```
devshgraphicsprogramming, line 7: kick this to common.hlsl

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+	std::array<core::smart_refctd_ptr<IGPUImage>, IMAGE_CNT> images;
+
+	static constexpr uint32_t FRAMES_IN_FLIGHT = 3u;
+	smart_refctd_ptr<video::IGPUCommandPool> commandPools[FRAMES_IN_FLIGHT];
+
+	static constexpr size_t CHANEL_CNT = 3;
+	static constexpr size_t VAL_PER_CHANEL_CNT = 256;
+	static constexpr size_t HISTOGRAM_SIZE = CHANEL_CNT * VAL_PER_CHANEL_CNT;
+	static constexpr size_t HISTOGRAM_BYTE_SIZE = HISTOGRAM_SIZE * sizeof(uint32_t);
+	static constexpr size_t COMBINED_HISTOGRAM_BUFFER_BYTE_SIZE = HISTOGRAM_BYTE_SIZE * FRAMES_IN_FLIGHT;
```
devshgraphicsprogramming, line 120: kick this to common.hlsl

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+		if (!asset_base_t::onAppInitialized(std::move(system)))
+			return false;
+
+		constexpr size_t TIMELINE_SEMAPHORE_STARTING_VALUE = 0;
+		m_imagesLoadedSemaphore = m_device->createSemaphore(TIMELINE_SEMAPHORE_STARTING_VALUE);
+		m_imagesProcessedSemaphore = m_device->createSemaphore(TIMELINE_SEMAPHORE_STARTING_VALUE);
+		m_histogramSavedSeamphore = m_device->createSemaphore(TIMELINE_SEMAPHORE_STARTING_VALUE);
+
+		// TODO: create/initialize array of atomic pointers to IGPUImage* and IGPUBuffer* to hold results
+		// no need i think
```
devshgraphicsprogramming, line 58: your `std::array<core::smart_refctd_ptr<IGPUImage>, IMAGE_CNT>` is fine, but whether the pointers need to be atomic depends on how you sync between `loadImages` and `calculateHistogram`

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+	static constexpr size_t CHANEL_CNT = 3;
+	static constexpr size_t VAL_PER_CHANEL_CNT = 256;
+	static constexpr size_t HISTOGRAM_SIZE = CHANEL_CNT * VAL_PER_CHANEL_CNT;
+	static constexpr size_t HISTOGRAM_BYTE_SIZE = HISTOGRAM_SIZE * sizeof(uint32_t);
+	static constexpr size_t COMBINED_HISTOGRAM_BUFFER_BYTE_SIZE = HISTOGRAM_BYTE_SIZE * FRAMES_IN_FLIGHT;
+	smart_refctd_ptr<IGPUBuffer> histogramBuffer = nullptr;
+	nbl::video::IDeviceMemoryAllocator::SAllocation m_histogramBufferAllocation = {};
+	// TODO: make sure ranges are ok
+	std::array<ILogicalDevice::MappedMemoryRange, FRAMES_IN_FLIGHT> m_histogramBufferMemoryRanges;
+	uint32_t* m_histogramBufferMemPtrs[3];
```
devshgraphicsprogramming, line 125: don't thing you need to keep `m_histogramBufferAllocation ` and `m_histogramBufferMemPtrs` around for all threads

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+
+	std::mutex assetManagerMutex;
```
devshgraphicsprogramming, line 127: AFAIK, asset manager doesn't need a mutex

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+
+	void loadImages()
+	{
+		IAssetLoader::SAssetLoadParams lp;
+		lp.logger = m_logger.get();
+
+		// LOAD CPU IMAGES
+		core::smart_refctd_ptr<ICPUImage> cpuImages[IMAGE_CNT];
+		for(uint32_t i = 0; i < IMAGE_CNT; ++i)
+			cpuImages[i] = loadFistAssetInBundle<ICPUImage>(imagesToLoad[i]);
```
devshgraphicsprogramming, line 137: thats actually the thing that takes the most time and RAM, move it  to the beginning of the for-loop

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
 
-	public:
-		// Yay thanks to multiple inheritance we cannot forward ctors anymore
-		StagingAndMultipleQueuesApp(const path& _localInputCWD, const path& _localOutputCWD, const path& _sharedInputCWD, const path& _sharedOutputCWD) :
-			system::IApplicationFramework(_localInputCWD,_localOutputCWD,_sharedInputCWD,_sharedOutputCWD) {}
+		auto transferUpQueue = getTransferUpQueue();
+		// TODO: i want to use IGPUCommandPool::CREATE_FLAGS::NONE but `updateImageViaStagingBuffer` requires command buffers to be resetable
+		const core::bitflag<IGPUCommandPool::CREATE_FLAGS> commandPoolFlags = static_cast<IGPUCommandPool::CREATE_FLAGS>(IGPUCommandPool::CREATE_FLAGS::RESET_COMMAND_BUFFER_BIT);
```
devshgraphicsprogramming, line 141: yes, indeed correct, thats how it needs to stay, btw you don't need that `static_cast` its redundant

devshgraphicsprogramming: but that means you only need one command-pool

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+			// TODO: make sure to use the non-blocking variant of `IUtilities` to control the `IQueue::SSubmitInfo::signalSemaphores`
+
+		for (uint32_t imageIdx = 0; imageIdx < IMAGE_CNT; ++imageIdx)
 		{
-			// Remember to call the base class initialization!
-			if (!device_base_t::onAppInitialized(std::move(system)))
-				return false;
-			if (!asset_base_t::onAppInitialized(std::move(system)))
-				return false;
+			const size_t resourceIdx = imageIdx % FRAMES_IN_FLIGHT;
```
devshgraphicsprogramming, line 160: you want to load a single image here

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
-			// - main thread grabs `IGPUImage` from a queue and acquires ownership [if necessary] performs all setup to launch a dispatch hands off a histogram buffer
+			auto isResourceReused = waitForResourceAvailability(m_imagesLoadedSemaphore.get(), imageIdx);
+			if(isResourceReused)
+				cmdBuff->reset(IGPUCommandBuffer::RESET_FLAGS::NONE);
 
-			return true;
+			IGPUImage::SCreationParams imgParams;
+			imgParams.type = IImage::E_TYPE::ET_2D;
+			imgParams.extent.height = cpuImages[0]->getCreationParameters().extent.height;
+			imgParams.extent.width = cpuImages[0]->getCreationParameters().extent.height;
```
devshgraphicsprogramming, line 171: typo, asigning `height` to `width`

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+			imgParams.extent.depth = cpuImages[0]->getCreationParameters().extent.depth;
+			IPhysicalDevice::SImageFormatPromotionRequest formatPromotionRequest;
+			IPhysicalDevice::SFormatImageUsages::SUsage usage;
+			usage.sampledImage = 1;
+			usage.transferDst = 1;
+			formatPromotionRequest.usages = usage;
+			// why my images are VK_FORMAT_R8_SRGB?
```
devshgraphicsprogramming, line 178: what do you mean?

devshgraphicsprogramming: ok I see, the format promotion fails?

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+			std::array familyIndices = { getTransferUpQueue()->getFamilyIndex(), getComputeQueue()->getFamilyIndex() };
+			imgParams.queueFamilyIndexCount = familyIndices.size();
+			imgParams.queueFamilyIndices = familyIndices.data();
+			imgParams.preinitialized = false;
+
+			images[imageIdx] = m_device->createImage(std::move(imgParams));
+			auto imageAllocation = m_device->allocate(images[imageIdx]->getMemoryReqs(), images[imageIdx].get(), IDeviceMemoryAllocation::EMAF_NONE);
+
+			if (!imageAllocation.isValid())
+				logFailAndTerminate("Failed to allocate Device Memory compatible with our GPU Buffer!\n");
```
devshgraphicsprogramming, line 196: correct message -> image not buffer

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+				imgSubresourceRange.aspectMask = IImage::E_ASPECT_FLAGS::EAF_COLOR_BIT;
+				imgSubresourceRange.baseMipLevel = 0u;
+				imgSubresourceRange.baseArrayLayer = 0u;
+				imgSubresourceRange.levelCount = 1;
+				imgSubresourceRange.layerCount = 1u;
+
+				imageLayoutTransitionBarrier0.barrier.dep.srcAccessMask = ACCESS_FLAGS::NONE;
+				imageLayoutTransitionBarrier0.barrier.dep.dstAccessMask = ACCESS_FLAGS::MEMORY_WRITE_BITS;
+				imageLayoutTransitionBarrier0.barrier.dep.srcStageMask = PIPELINE_STAGE_FLAGS::NONE;
+				imageLayoutTransitionBarrier0.barrier.dep.dstStageMask = PIPELINE_STAGE_FLAGS::ALL_TRANSFER_BITS; //TODO: COPY_BIT?
```
devshgraphicsprogramming, line 212: yes copy-bit

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+				imageLayoutTransitionBarrier0.oldLayout = asset::IImage::LAYOUT::UNDEFINED;
+				imageLayoutTransitionBarrier0.newLayout = asset::IImage::LAYOUT::TRANSFER_DST_OPTIMAL; // TODO: use more suitable layout
```
devshgraphicsprogramming, line 214: that is the correct layout

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+				imageLayoutTransitionBarrier0.image = images[imageIdx].get();
+				imageLayoutTransitionBarrier0.subresourceRange = imgSubresourceRange;
+
+				imageLayoutTransitionBarrier1.barrier.dep.srcAccessMask = ACCESS_FLAGS::MEMORY_WRITE_BITS; 
+				imageLayoutTransitionBarrier1.barrier.dep.dstAccessMask = ACCESS_FLAGS::NONE;
+				imageLayoutTransitionBarrier1.barrier.dep.srcStageMask = PIPELINE_STAGE_FLAGS::ALL_TRANSFER_BITS; //TODO: COPY_BIT?
+				imageLayoutTransitionBarrier1.barrier.dep.dstStageMask = PIPELINE_STAGE_FLAGS::NONE;
```
devshgraphicsprogramming, line 221: src state is COPY_BIT

the dest stage and acces can be NONE, but you need to comment that its because of the semaphore signal coming right after

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+				imageLayoutTransitionBarrier1.oldLayout = asset::IImage::LAYOUT::TRANSFER_DST_OPTIMAL;
+				imageLayoutTransitionBarrier1.newLayout = asset::IImage::LAYOUT::GENERAL; // TODO: use more suitable layout
```
devshgraphicsprogramming, line 223: SHADER_READ_ONLY since you'll be sampling

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+				imageLayoutTransitionBarrier1.image = images[imageIdx].get();
+				imageLayoutTransitionBarrier1.subresourceRange = imgSubresourceRange;
+			}
+
+			core::smart_refctd_ptr<ISemaphore> imgFillSemaphore = m_device->createSemaphore(0); // TODO: don't create semaphore every iteration
```
devshgraphicsprogramming, line 228: you can create one outside the loop and just reuse it, you need to keep track of the next signal value as well

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+				.value = 1,
+				.stageMask = PIPELINE_STAGE_FLAGS::ALL_TRANSFER_BITS
+			};
+
+			IQueue::SSubmitInfo imgFillSubmitInfo = {
+				.waitSemaphores = {&imgFillSemaphoreWaitInfo, 1},
+				.commandBuffers = {&imgFillCmdBuffInfo, 1}
+			};
+
+			transferUpQueue->submit({ &imgFillSubmitInfo, 1 });
```
devshgraphicsprogramming, line 242: why are you doing a submit of an empty (freshly reset commandbuffer), just to signal a temporary semaphore to 1 !?

devshgraphicsprogramming: especially since that your `intendedSubmit` uses a `value` of 0 for its scratch semaphore

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+				.stageMask = PIPELINE_STAGE_FLAGS::ALL_TRANSFER_BITS
+			};
+			SIntendedSubmitInfo intendedSubmit = {
+				.frontHalf = {.queue = transferUpQueue, .waitSemaphores = {}, .commandBuffers = {&imgFillCmdBuffInfo, 1}}, .signalSemaphores = {&imgFillSemaphoreInfo, 1}
+			};
+			
+			cmdBuff->begin(IGPUCommandBuffer::USAGE::ONE_TIME_SUBMIT_BIT);
+
+			IGPUCommandBuffer::SPipelineBarrierDependencyInfo pplnBarrierDepInfo0;
+			pplnBarrierDepInfo0.imgBarriers = std::span(&imageLayoutTransitionBarrier0, &imageLayoutTransitionBarrier0 + 1);
```
devshgraphicsprogramming, line 257: easier shorthand is `= {&imageLayoutTransitionBarrier0, 1};`

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+				logFailAndTerminate("Failed to issue barrier!\n");
+
+			cmdBuff->end();
+
+			IQueue::SSubmitInfo submitInfo[1];
+			IQueue::SSubmitInfo::SCommandBufferInfo cmdBuffSubmitInfo[] = { {cmdBuff.get()} };
+			IQueue::SSubmitInfo::SSemaphoreInfo signalSemaphoreSubmitInfo[] = { { .semaphore = m_imagesLoadedSemaphore.get(), .value = imageIdx+1, .stageMask = PIPELINE_STAGE_FLAGS::ALL_COMMANDS_BITS } };
+			submitInfo[0].commandBuffers = cmdBuffSubmitInfo;
+			submitInfo[0].signalSemaphores = signalSemaphoreSubmitInfo;
+			getTransferUpQueue()->submit(submitInfo);
```
devshgraphicsprogramming, line 282: here's the thing, to use `SIntededSubmit` properly you need to cast (implicit convert) it to `SSubmitInfo`

then you can mofidy the signal span slightly to also signal the `m_imagesLoadedSemaphore`

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+		}
+
+		for (uint32_t imageHistogramIdx = 0; imageHistogramIdx < IMAGE_CNT; ++imageHistogramIdx)
+		{
+			waitForPreviousStep(m_imagesProcessedSemaphore.get(), imageHistogramIdx + 1);
+
+			const uint32_t resourceIdx = imageHistogramIdx % FRAMES_IN_FLIGHT;
+			uint32_t* histogramBuff = m_histogramBufferMemPtrs[resourceIdx];
+
+			waitForResourceAvailability(m_histogramSavedSeamphore.get(), imageHistogramIdx);
```
devshgraphicsprogramming, line 497: this is redundant, because its the histogram computing dispatch (image processing) needs to wait for the histogram output range to stop being used

devshgraphicsprogramming: and because you're wiping the histogram it needs to wait on `m_histogramSavedSemaphore` before the dispatch

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+			m_histogramBufferMemPtrs[1] = m_histogramBufferMemPtrs[0] + HISTOGRAM_SIZE;
+			m_histogramBufferMemPtrs[2] = m_histogramBufferMemPtrs[1] + HISTOGRAM_SIZE;
 		}
 
-};
+		// PROCESS IMAGES
+		for (uint32_t imageToProcessId = 0; imageToProcessId < IMAGE_CNT; imageToProcessId++)
+		{
+			// TODO: wait for atomic instead? will it be even valid to do? we need to wait for gpu to finish anyway..
+			waitForPreviousStep(m_imagesLoadedSemaphore.get(), imageToProcessId + 1);
```
devshgraphicsprogramming, line 409: atomic would have been LoadThread->ProcessThread

what you have now is LoadThread->UploadQueue->ProcessThread->ProcessQueue

What you want is LoadThread->UploadQueue->ProcessQueue


you need a `m_imagesLoadedSemaphore, imageToProcessId+1, COPY_BIT` wait in the `submitInfo.waitSemaphores`

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+			IGPUImageView::SCreationParams params{};
+			params.viewType = IImageView<IGPUImage>::ET_2D;
+			params.image = images[imageToProcessId];
+			params.format = images[imageToProcessId]->getCreationParameters().format;
+			params.subresourceRange.aspectMask = IImage::E_ASPECT_FLAGS::EAF_COLOR_BIT;
+			params.subresourceRange.layerCount = images[resourceIdx]->getCreationParameters().arrayLayers;
+
+			IGPUSampler::SParams samplerParams;
+			samplerParams.AnisotropicFilter = false;
+			core::smart_refctd_ptr<IGPUSampler> sampler = m_device->createSampler(samplerParams);
```
devshgraphicsprogramming, line 430: create it up-front and reuse

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+
+			imgInfo.desc = m_device->createImageView(std::move(params));
+			if (!imgInfo.desc)
+				logFailAndTerminate("Couldn't create descriptor.");
+			imgInfo.info.image = { .sampler = sampler, .imageLayout = IImage::LAYOUT::GENERAL };
```
devshgraphicsprogramming, line 435: readonly layout would be better

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+
+			IGPUDescriptorSet::SDescriptorInfo bufInfo;
+			bufInfo.desc = smart_refctd_ptr(histogramBuffer);
+			bufInfo.info.buffer = { .offset = 0u, .size = histogramBuffer->getSize() };
```
devshgraphicsprogramming, line 439: you can write the histogram buffer up-front into all descriptor sets

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+			computeQueue->startCapture();
+			cmdBuff->begin(IGPUCommandBuffer::USAGE::NONE);
+			cmdBuff->beginDebugMarker("My Compute Dispatch", core::vectorSIMDf(0, 1, 0, 1));
+			cmdBuff->bindComputePipeline(pipeline.get());
+
+			cmdBuff->bindDescriptorSets(nbl::asset::EPBP_COMPUTE, pplnLayout.get(), 0, 1, &descSets[resourceIdx].get());
+
+			const auto imageExtent = images[imageToProcessId]->getCreationParameters().extent;
+			const uint32_t wgCntX = imageExtent.width / WorkgroupSizeX;
+			const uint32_t wgCntY = imageExtent.height / WorkgroupSizeY;
```
devshgraphicsprogramming, line 456: you need to round up by dividing, either `(x-1)/d+1` or `(x+d-1)/d`

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+			cmdBuff->endDebugMarker();
+			cmdBuff->end();
+
+			IQueue::SSubmitInfo submitInfo[1];
+			IQueue::SSubmitInfo::SCommandBufferInfo cmdBuffSubmitInfo[] = { {cmdBuff.get()} };
+			IQueue::SSubmitInfo::SSemaphoreInfo signalSemaphoreSubmitInfo[] = { {.semaphore = m_imagesProcessedSemaphore.get(), .value = imageToProcessId+1, .stageMask = PIPELINE_STAGE_FLAGS::ALL_COMMANDS_BITS } };
+			submitInfo[0].commandBuffers = cmdBuffSubmitInfo;
+			submitInfo[0].signalSemaphores = signalSemaphoreSubmitInfo;
+
+			waitForResourceAvailability(m_histogramSavedSeamphore.get(), imageToProcessId);
```
devshgraphicsprogramming, line 473: you actually want to wait on `m_histogramSavedSeamphore` FRAMES_IN_FLIGHT behind

devshgraphicsprogramming: so you'd have 2 waits [GPU waits] in the submit:
- this image is loaded `m_imagesLoadedSemaphore` 
- the histogram offset you'll be writing to is no longer being used `m_histogramSavedSeamphore+1-FRAMES_IN_FLIGHT`

both wait stages would be COMPUTE

aside from that you'd be waiting on with the CPU before going through with the commandbuffer reset, so on `m_imagesProcessedSemaphore+1-FRAMES_IN_FLIGHT`

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+
+			cmdBuff->pushConstants(pplnLayout.get(), IShader::E_SHADER_STAGE::ESS_COMPUTE, 0u, sizeof(PushConstants), &constants);
+			cmdBuff->dispatch(wgCntX, wgCntY, 1);
+
+			cmdBuff->endDebugMarker();
+			cmdBuff->end();
+
+			IQueue::SSubmitInfo submitInfo[1];
+			IQueue::SSubmitInfo::SCommandBufferInfo cmdBuffSubmitInfo[] = { {cmdBuff.get()} };
+			IQueue::SSubmitInfo::SSemaphoreInfo signalSemaphoreSubmitInfo[] = { {.semaphore = m_imagesProcessedSemaphore.get(), .value = imageToProcessId+1, .stageMask = PIPELINE_STAGE_FLAGS::ALL_COMMANDS_BITS } };
```
devshgraphicsprogramming, line 469: stage would be compute, you signal after compute is done

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+
+			m_histogramBufferMemoryRanges[0] = ILogicalDevice::MappedMemoryRange(histogramBuffer->getBoundMemory().memory, 0, HISTOGRAM_BYTE_SIZE);
+			m_histogramBufferMemoryRanges[1] = ILogicalDevice::MappedMemoryRange(histogramBuffer->getBoundMemory().memory, HISTOGRAM_BYTE_SIZE, HISTOGRAM_BYTE_SIZE);
+			m_histogramBufferMemoryRanges[2] = ILogicalDevice::MappedMemoryRange(histogramBuffer->getBoundMemory().memory, HISTOGRAM_BYTE_SIZE * 2, HISTOGRAM_BYTE_SIZE);
+
+			m_histogramBufferMemPtrs[0] = static_cast<uint32_t*>(m_histogramBufferAllocation.memory->map(memoryRange, IDeviceMemoryAllocation::EMCAF_READ_AND_WRITE));
+			if (!m_histogramBufferMemPtrs[0])
+				logFailAndTerminate("Failed to map the Device Memory!\n");
+			m_histogramBufferMemPtrs[1] = m_histogramBufferMemPtrs[0] + HISTOGRAM_SIZE;
+			m_histogramBufferMemPtrs[2] = m_histogramBufferMemPtrs[1] + HISTOGRAM_SIZE;
```
devshgraphicsprogramming, line 402: they're good, could be local vars though

07_StagingAndMultipleQueues/main.cpp
devshgraphicsprogramming: The `m_imagesLoadedSemaphore` you signal from and wait at ALL (because of the layout transition)
and `imgFillSemaphore` you signal from and wait at TRANSFER

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
-		// Ideally don't want to have to 
-		SPhysicalDeviceFeatures getPreferredDeviceFeatures() const override
+		// CREATE AND MAP HISTOGRAM BUFFER
 		{
-			auto retval = device_base_t::getPreferredDeviceFeatures();
-			retval.shaderStorageImageReadWithoutFormat = true;
-			return retval;
+			IGPUBuffer::SCreationParams gpuBufCreationParams;
+			gpuBufCreationParams.size = COMBINED_HISTOGRAM_BUFFER_BYTE_SIZE;
+			gpuBufCreationParams.usage = IGPUBuffer::E_USAGE_FLAGS::EUF_TRANSFER_DST_BIT | IGPUBuffer::E_USAGE_FLAGS::EUF_TRANSFER_SRC_BIT | IGPUBuffer::E_USAGE_FLAGS::EUF_STORAGE_BUFFER_BIT;
```
devshgraphicsprogramming, line 376: transfer usages are for copy commands, if you just map, you can do just STORAGE

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+			if (!cmdBuff->pipelineBarrier(E_DEPENDENCY_FLAGS::EDF_NONE, pplnBarrierDepInfo0))
+				logFailAndTerminate("Failed to issue barrier!\n");
+
+			transferUpQueue->startCapture();
+			bool uploaded = m_utils->updateImageViaStagingBuffer(
+				intendedSubmit, cpuImages[imageIdx]->getBuffer(), cpuImages[imageIdx]->getCreationParameters().format,
+				images[imageIdx].get(), IImage::LAYOUT::TRANSFER_DST_OPTIMAL, cpuImages[imageIdx]->getRegions()
+			);
+			if (!uploaded)
+				logFailAndTerminate("Couldn't update image data.\n");
```
devshgraphicsprogramming, line 267: you can also zero-out/nullify the cpuimage right after

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -2,78 +2,580 @@
...
+		std::array<std::ofstream, IMAGE_CNT> files;
+		for (auto& file : files)
+		{
+			static uint32_t i = 0u;
+			file.open("histogram_" + std::to_string(i++) + ".csv", std::ios::out | std::ios::trunc);
+		}
+
+		for (uint32_t imageHistogramIdx = 0; imageHistogramIdx < IMAGE_CNT; ++imageHistogramIdx)
+		{
+			waitForPreviousStep(m_imagesProcessedSemaphore.get(), imageHistogramIdx + 1);
```
devshgraphicsprogramming, line 492: as soon as the CPU returns, you can nullify/zero-out the gpuimage

07_StagingAndMultipleQueues/main.cpp
devshgraphicsprogramming: you're not even touching it from multiple threads XD

07_StagingAndMultipleQueues/main.cpp
Przemog1: i do use it in both `saveHistograms` and `calculateHistograms`

07_StagingAndMultipleQueues/main.cpp
Przemog1: seems like it

07_StagingAndMultipleQueues/main.cpp
devshgraphicsprogramming: why does `calculateHistograms` need to know the `void*` pointer to the mapping?

Przemog1: idea is to create and map the buffer in `calculateHistograms` to avoid additional synchronization between `calculateHistograms` and `saveHistograms`.
I could map pointers in `saveHistograms` and keep these pointers local there but then i would have to wait for `calculateHistograms` to create the buffer first.
So current solution seems ok to me..

07_StagingAndMultipleQueues/main.cpp
devshgraphicsprogramming: ok I'll use this as a test and fix

07_StagingAndMultipleQueues/main.cpp
Przemog1: but here: https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/blob/vulkan_1_3_example_07_fix/07_StagingAndMultipleQueues/main.cpp#L419-L426 i'm using a GPU image in CPU to update descriptor set write, so i'm touching the loaded images with CPU no?

07_StagingAndMultipleQueues/main.cpp
Przemog1: ok, it was my fault, i haven't set an `originalFormat` field of the `SImageFormatPromotionRequest` struct

07_StagingAndMultipleQueues/main.cpp
devshgraphicsprogramming: no, you're touching the Descriptor set with CPU

07_StagingAndMultipleQueues/main.cpp
```diff
@@ -13,67 +13,548 @@ using namespace system;
...
+				.createFlags = IGPUDescriptorSetLayout::SBinding::E_CREATE_FLAGS::ECF_NONE,
+				.stageFlags = IGPUShader::E_SHADER_STAGE::ESS_COMPUTE,
+				.count = 1,
+				.samplers = nullptr
+			}
+		};
+		smart_refctd_ptr<IGPUDescriptorSetLayout> dsLayout[1] = { m_device->createDescriptorSetLayout(bindings) };
+		if (!dsLayout[0])
+			logFailAndTerminate("Failed to create a Descriptor Layout!\n");
+		smart_refctd_ptr<nbl::video::IDescriptorPool> descPools[FRAMES_IN_FLIGHT] = { // TODO: only one desc pool?
```
devshgraphicsprogramming, line 288: yes thats okay

PR #102, opened 2024-03-19
BubbleSort demo

Fletterio:
Adds an unnumbered XX_BubbleSort example to my task's branch


XX_BubbleSort/app_resources/shader.comp.hlsl
```diff
@@ -0,0 +1,33 @@
+#include "common.hlsl"
+
+// just a small test
+#include "nbl/builtin/hlsl/jit/device_capabilities.hlsl"
+
+[[vk::push_constant]] PushConstantData pushConstants;
+
+// does absolutely nothing, a later example will show how it gets used
+template<typename capability_traits=nbl::hlsl::jit::device_capabilities_traits>
+void dummyTraitTest() {}
```
devshgraphicsprogramming, line 10: remove stuff like this

XX_BubbleSort/app_resources/shader.comp.hlsl
```diff
@@ -0,0 +1,33 @@
...
+	const static uint32_t OthersToTest = 15;
+	[[unroll(OthersToTest)]]
+	for (uint32_t i=0; i<OthersToTest; i++)
+	{
+		const uint32_t offset = rng() % pushConstants.dataElementCount;
+		const input_t other = vk::RawBufferLoad<input_t>(pushConstants.inputAddress+sizeof(input_t)*offset);
+		acc = min(length(other-self),acc);
+	}
+	vk::RawBufferStore<float32_t>(pushConstants.outputAddress+sizeof(float32_t)*ID.x,acc);
+}
```
devshgraphicsprogramming, line 33: looks like thats just the old shader from ex 02, maybe delete it if you're not using it?

XX_BubbleSort/app_resources/shader.comp.hlsl
devshgraphicsprogramming: waiit, you're not using this shader at all

XX_BubbleSort/main.cpp
```diff
@@ -0,0 +1,449 @@
...
+
+							uint32_t aux = 0;
+
+							if (buff[N+1] > buff[N+2]) {
+								aux = buff[N+1];
+								buff[N+1] = buff[N+2];
+								buff[N+2] = aux;
+							}
+						}
+					)===";
```
devshgraphicsprogramming, line 169: Please use `app_resources`, don't embed shader source in the executable, because:
1. we don't get any syntax highlighting
2. we have a special Cmake option to not embed the resources in a DLL, which makes it much faster to iterate (no recompile of EXE needed)

XX_BubbleSort/main.cpp
```diff
@@ -0,0 +1,449 @@
...
+					.stageFlags = IGPUShader::ESS_COMPUTE,
+					.count = 1,
+					.samplers = nullptr // irrelevant for a buffer
+				}
+			};
+			smart_refctd_ptr<IGPUDescriptorSetLayout> dsLayout = device->createDescriptorSetLayout(bindings);
+			if (!dsLayout)
+				return logFail("Failed to create a Descriptor Layout!\n");
+
+			const nbl::asset::SPushConstantRange pcRange = { .stageFlags = IShader::ESS_COMPUTE,.offset = 0,.size = sizeof(PushConstantData) };
```
devshgraphicsprogramming, line 217: Renderdoc now supports BDA, so please use BDA and stick the buffer address in the Push Constant, this way you don't need to create descriptor sets, etc.

XX_BubbleSort/main.cpp
```diff
@@ -0,0 +1,449 @@
...
+
+			// Nabla actually has facilities for SPIR-V Reflection and "guessing" pipeline layouts for a given SPIR-V which we'll cover in a different example
+			smart_refctd_ptr<nbl::video::IGPUPipelineLayout> pplnLayout = device->createPipelineLayout({ &pcRange, 1 }, smart_refctd_ptr(dsLayout));
+			if (!pplnLayout)
+				return logFail("Failed to create a Pipeline Layout!\n");
+
+			// We use strong typing on the pipelines (Compute, Graphics, Mesh, RT), since there's no reason to polymorphically switch between different pipelines
+			smart_refctd_ptr<nbl::video::IGPUComputePipeline> evenPipeline;
+			smart_refctd_ptr<nbl::video::IGPUComputePipeline> oddPipeline;
```
devshgraphicsprogramming, line 226: Also you can use `pushConstants` to control odd vs. even, no need for separate pipelines.

XX_BubbleSort/main.cpp
```diff
@@ -0,0 +1,449 @@
...
+								return;
+
+							uint32_t aux = 0;
+
+							if (buff[N] > buff[N+1]) {
+								aux = buff[N];
+								buff[N] = buff[N+1];
+								buff[N+1] = aux;
+							}
+						}
```
devshgraphicsprogramming, line 139: if you were to refactor this into an `ext` I'd wrap the entire `main` into a struct and template on the type you want to sort, i.e.
```cpp
namespace nbl 
{
namespace hlsl 
{

struct bubble_sort
{
   template<typename Accessor, typename Comparator>
   static void __call(NBL_ARG_REF(Accessor) accessor, NBL_ARG_REF(Comparator) comp)
   {
   }
};

}
}
```

Note: because HLSL doesn't have references/pointers we use a "stateful Accessor pattern" (similar to a C++ STL iterator) see the workgroup prefix sum (ex 22) to see what we mean.

XX_BubbleSort/main.cpp
```diff
@@ -0,0 +1,449 @@
...
+						break;
+					}
+			}
+			if (!physDev)
+				return logFail("Failed to find any Physical Devices with Compute capable Queue Families!");
+
+			// logical devices need to be created form physical devices which will actually let us create vulkan objects and use the physical device
+			device = physDev->createLogicalDevice(std::move(params));
+			if (!device)
+				return logFail("Failed to create a Logical Device!");
```
devshgraphicsprogramming, line 95: Hello Compute was special here, because its the 2nd example, so it did more stuff "from the ground up"

You chouls use a more complex base class, like `MonoDeviceApplication` or `BasicMultiQueue`... then you can skip all this queue choice, device creation code... see example 05

devshgraphicsprogramming:
Reopen the PR against the `master` branch

PR #105, opened 2024-03-26
Counting Sort Example

nipunG314:


CountingSort/app_resources/shader.comp.hlsl
```diff
@@ -0,0 +1,22 @@
...
+        return;
+
+    int count = 0;
+    for (uint32_t index = 0; index < pushConstants.dataElementCount; index++)
+    {
+        if (vk::RawBufferLoad(pushConstants.inputAddress + sizeof(uint32_t) * index) == ID.x)
+            count++;
+    }
+
+    vk::RawBufferStore(pushConstants.outputAddress + sizeof(uint32_t) * ID.x, count);
```
devshgraphicsprogramming, line 21: ok this is never gonna work

CountingSort/app_resources/prefix_sum_shader.comp.hlsl
```diff
@@ -0,0 +1,76 @@
...
+    uint32_t tid = nbl::hlsl::workgroup::SubgroupContiguousIndex();
+    sdata[tid] = 0;
+    uint32_t index = nbl::hlsl::glsl::gl_WorkGroupID().x * WorkgroupSize + tid;
+
+    nbl::hlsl::glsl::barrier();
+
+    if (index < pushConstants.dataElementCount)
+    {
+        uint32_t value = vk::RawBufferLoad(pushConstants.inputAddress + sizeof(uint32_t) * index);
+        nbl::hlsl::glsl::atomicAdd(sdata[value - pushConstants.minimum], (uint32_t) 1);
```
devshgraphicsprogramming, line 58: btw does using `uint16_t` with shared atomics require special extensions?

CountingSort/app_resources/prefix_sum_shader.comp.hlsl
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    nbl::hlsl::glsl::barrier();
+
+    //nbl::hlsl::workgroup::ballot < decltype(ballotAccessor) > (bool(sdata[tid]), ballotAccessor);
+    //ballotAccessor.workgroupExecutionAndMemoryBarrier();
+    //uint32_t dest_value = nbl::hlsl::workgroup::ballotExclusiveBitCount < WorkgroupSize, decltype(ballotAccessor), decltype(arithmeticAccessor), nbl::hlsl::jit::device_capabilities > (ballotAccessor, arithmeticAccessor);
```
devshgraphicsprogramming, line 65: what were you trying to do with a ballot scan?

CountingSort/main.cpp
```diff
@@ -0,0 +1,382 @@
...
+		smart_refctd_ptr<IGPUComputePipeline> scatterPipeline;
+		{
+			layout = m_device->createPipelineLayout({ &pcRange,1 }, smart_refctd_ptr(dsLayout));
+			IGPUComputePipeline::SCreationParams params = {};
+			params.layout = layout.get();
+			params.shader.shader = prefixSumShader.get();
+			params.shader.entryPoint = "main";
+			params.shader.entries = nullptr;
+			params.shader.requireFullSubgroups = true;
+			params.shader.requiredSubgroupSize = static_cast<IGPUShader::SSpecInfo::SUBGROUP_SIZE>(5);
```
devshgraphicsprogramming, line 109: good, tis what we want

CountingSort/app_resources/scatter_shader.comp.hlsl
```diff
@@ -0,0 +1,19 @@
...
+[numthreads(WorkgroupSize, 1, 1)]
+void main(uint32_t3 ID : SV_GroupThreadID, uint32_t3 GroupID : SV_GroupID)
+{
+    uint32_t index = WorkgroupSize * GroupID.x + ID.x;
+
+    if (index < pushConstants.dataElementCount)
+    {
+        uint32_t value = vk::RawBufferLoad(pushConstants.inputAddress + sizeof(uint32_t) * index);
+        uint32_t address = nbl::hlsl::glsl::atomicAdd(scratch[value], (uint32_t) -1);
+        vk::RawBufferStore(pushConstants.outputAddress + sizeof(uint32_t) * (address - 1), value);
```
devshgraphicsprogramming, line 17: using an Exclusive Scan might make the code here prettier, less `-1` in the mix.

Btw to optimize, preload `scratch` into `groupshared`, atomics on shared memory are faster.

nipunG314: Wouldn't we have to atomicAdd to the vram buffer eventually to keep the histogram up to date?

devshgraphicsprogramming: good point, you can "re-perform" the histogramming into `smem` again, then `smem[value] = atomicAdd(scratch[value],smem[value])` to preload

obviously whats faster will depend on the ratio of elements processed per workgroup and the bucket count

CountingSort/app_resources/prefix_sum_shader.comp.hlsl
nipunG314: I have not enabled any special extensions with my vulkan installation.

CountingSort/app_resources/scatter_shader.comp.hlsl
nipunG314: I didn't understand what you meant by "re-perform" the histogramming

devshgraphicsprogramming: re-build the workgroup-local histogram, same way you do in the first compute dispatch before you prefix sum and accumulate globally

CountingSort/app_resources/prefix_sum_shader.comp.hlsl
devshgraphicsprogramming: This
```cpp
spirv::atomicAdd<uint16_t>(sdata[...], /*spv::Workgroup*/2, /*spv::MaskNone*/0, value);
```
would appear to compile https://godbolt.org/z/cs66bh5bo

I'd need to change a few things in my glsl_compat headers though

CountingSort/app_resources/scatter_shader.comp.hlsl
nipunG314: Done

CountingSort/app_resources/prefix_sum_shader.comp.hlsl
```diff
@@ -60,22 +60,28 @@ void main(uint32_t3 ID : SV_GroupThreadID, uint32_t3 GroupID : SV_GroupID)
...
         
         if ((tid == WorkgroupSize - 1) && i < (BucketsPerThread - 1))
-            sdata[WorkgroupSize * (i + 1)] += sum;
-    
+        {
+            scan_sum = sum + sdata[WorkgroupSize * i + tid];
+            sdata[WorkgroupSize * (i + 1)] += scan_sum;
+        }
+
         arithmeticAccessor.workgroupExecutionAndMemoryBarrier();
```
devshgraphicsprogramming, line 85: when writing loops like this you want to reorder them so the barrier appears first (last barrier is usually useless in a classical/naive loop)

CountingSort/app_resources/prefix_sum_shader.comp.hlsl
```diff
@@ -60,22 +60,28 @@ void main(uint32_t3 ID : SV_GroupThreadID, uint32_t3 GroupID : SV_GroupID)
...
         
         if ((tid == WorkgroupSize - 1) && i < (BucketsPerThread - 1))
-            sdata[WorkgroupSize * (i + 1)] += sum;
-    
+        {
+            scan_sum = sum + sdata[WorkgroupSize * i + tid];
+            sdata[WorkgroupSize * (i + 1)] += scan_sum;
+        }
+
```
devshgraphicsprogramming, line 84: this is quite convoluted, you'd need to revert to `inclusive_scan` though, then you can do this
```cpp
uint32_t lastResult;
{
   uint32_t scanInput = scratch[WorkgroupSize * i + tid];
   if (i>0) // a value added to the first entry of the scan will propagate throughout
      scanInput = lastResult;
   acccessor.barrier();
   uint32_t result = inclusive_scan(scanInput,accessor);
   nbl::hlsl::glsl::atomicAdd(scratch[WorkgroupSize * i + tid], result);
   lastResult = result;
}
```

if you want an exclusive scan in the scatter pass, you can just read the `scratch` at offset +1

CountingSort/app_resources/scatter_shader.comp.hlsl
```diff
@@ -14,16 +18,34 @@ uint32_t3 nbl::hlsl::glsl::gl_WorkGroupSize()
...
     uint32_t index = (nbl::hlsl::glsl::gl_WorkGroupID().x * WorkgroupSize + tid) * pushConstants.elementsPerWT;
 
     nbl::hlsl::glsl::barrier();
     
+    [unroll]
     for (int i = 0; i < pushConstants.elementsPerWT; i++)
     {
         if (index + i >= pushConstants.dataElementCount)
             break;
         uint32_t value = vk::RawBufferLoad(pushConstants.inputAddress + sizeof(uint32_t) * (index + i));
```
devshgraphicsprogramming, line 34: learn about coalesced loads, when indexing elements to load you should have consecutive `InvocationIndex` load consecutive elements, your current sceheme is is really bad for cachelines

workgroup > subgroup > invocation

if you want to repeat elements, they should repeat at the workgroup granularity (i.e. isntead of invocation 0 processing element 0 and 1, they should process element 0, WorkgroupSize)

CountingSort/app_resources/scatter_shader.comp.hlsl
```diff
@@ -14,16 +18,34 @@ uint32_t3 nbl::hlsl::glsl::gl_WorkGroupSize()
...
+    [unroll]
+    for (int i = 0; i < BucketsPerThread; i++)
+    {
+        uint32_t count = sdata[WorkgroupSize * i + tid];
+        sdata[WorkgroupSize * i + tid] = nbl::hlsl::glsl::atomicAdd(scratch[WorkgroupSize * i + tid], sdata[WorkgroupSize * i + tid]);
+        [unroll]
+        for (int j = 0; j < count; j++)
+        {
+            vk::RawBufferStore(pushConstants.outputAddress + sizeof(uint32_t) * (sdata[WorkgroupSize * i + tid] + j), WorkgroupSize * i + tid);
+        }
```
devshgraphicsprogramming, line 49: this is bad work-balancing, you still want 1 thread to perform `elementsPerWT` stores

I know you're trying to avoid a second `vk::RawBufferLoad` but you'd usually sort Key,Value SoA pairs, so they histogramming passes would only ready Key, while the final scatter here would touch `Value` instead

you need a fourth loop which does Load/Store while atomicAdd(-1) on the `sdata` to figure out the destination index

CountingSort/app_resources/prefix_sum_shader.comp.hlsl
```diff
@@ -0,0 +1,16 @@
...
+uint32_t3 nbl::hlsl::glsl::gl_WorkGroupSize()
+{
+    return uint32_t3(WorkgroupSize, 1, 1);
+}
+
+[numthreads(WorkgroupSize,1,1)]
+void main(uint32_t3 ID : SV_GroupThreadID, uint32_t3 GroupID : SV_GroupID)
+{
+    nbl::hlsl::sort::counting < bda::__ptr < uint32_t >, bda::__ptr < uint32_t >, bda::__ptr < uint32_t > > counter;
```
devshgraphicsprogramming, line 13: design should be slightly different, don't use BDA pointers directly, lets have Accessors

Accessor pattern is important because it lets you use any underlying resource via static polymorphism"
- bda pointers
- ssbos/ubos
- sampler buffers
- textures
- shared memory

nipunG314: You mean rewrite the histogram() and scatter() functions to be completely generic?

CountingSort/app_resources/prefix_sum_shader.comp.hlsl
nipunG314: Done

CountingSort/app_resources/prefix_sum_shader.comp.hlsl
nipunG314: Aren't we completely ignoring the sdata[] array that contains the data that needs to be summed & uploaded to scratch? And for i>0, it seems to ignore the values of the current iteration for the previous result.

CountingSort/app_resources/scatter_shader.comp.hlsl
nipunG314: Done

CountingSort/app_resources/scatter_shader.comp.hlsl
nipunG314: For the Key,Value pairs, are we supposed to sort by Key or Value? I thought we were supposed to sort by Value? Other than that, we do that now

CountingSort/app_resources/prefix_sum_shader.comp.hlsl
devshgraphicsprogramming: not functions.

Accessor is a concept, e.g.
```cpp
struct MyAccessor
{
    uint32_t get(uint64_t index);
    void set(uint64_t index, uint32_t value);
};
```

If you use a `MyAccessor` that has those methods, the utility should work.

CountingSort/app_resources/scatter_shader.comp.hlsl
devshgraphicsprogramming: You sort by Key obviously, its associated Value just moves with it.

So if key originally at offset i ends up at offset j, the value should too.

This is why you only histogram keys, but scatter both.

CountingSort/app_resources/scatter_shader.comp.hlsl
```diff
@@ -10,7 +11,13 @@ uint32_t3 nbl::hlsl::glsl::gl_WorkGroupSize()
...
-    nbl::hlsl::sort::counting < bda::__ptr < uint32_t >, bda::__ptr < uint32_t >, bda::__ptr < uint32_t >  > counter;
-    counter.init(pushData);
-    counter.scatter();
+    nbl::hlsl::sort::counting < bda::PtrAccessor<uint32_t>, bda::PtrAccessor<uint32_t>, bda::PtrAccessor<uint32_t> > counter;
+    counter.scatter(
+        bda::PtrAccessor<uint32_t>::createAccessor(pushData.inputKeyAddress),
+        bda::PtrAccessor<uint32_t>::createAccessor(pushData.inputValueAddress),
+        bda::PtrAccessor<uint32_t>::createAccessor(pushData.scratchAddress),
+        bda::PtrAccessor<uint32_t>::createAccessor(pushData.outputKeyAddress),
+        bda::PtrAccessor<uint32_t>::createAccessor(pushData.outputValueAddress),
```
devshgraphicsprogramming, line 20: I'd roll up the input/output accessor pairs into one, because they're accessors you can make a custom one that gets from a different address than sets

nipunG314: Done

CountingSort/app_resources/prefix_sum_shader.comp.hlsl
```diff
@@ -1,8 +1,39 @@
...
+        deref().store(value);
+    }
+
+    uint32_t atomicAdd(uint64_t index, uint32_t value)
+    {
+        return bda::__ptr < uint32_t > (addr + sizeof(uint32_t) * index).template
+        deref().atomicAdd(value);
+    }
+
+    uint64_t addr;
```
devshgraphicsprogramming, line 34: you want a separate source and dest address, so you can load and store in parallel when scattering

CountingSort/app_resources/scatter_shader.comp.hlsl
nipunG314: Done

CountingSort/app_resources/prefix_sum_shader.comp.hlsl
```diff
@@ -0,0 +1,53 @@
+#include "nbl/builtin/hlsl/bda/__ptr.hlsl"
+#include "nbl/builtin/hlsl/sort/counting.hlsl"
+#include "nbl/builtin/hlsl/sort/common.hlsl"
+
+[[vk::push_constant]] nbl::hlsl::sort::CountingPushData pushData;
+
+struct PtrAccessor
+{
+    static PtrAccessor createAccessor(uint64_t addr)
```
devshgraphicsprogramming, line 9: we have a convention of calling HLSL-pretend-constructors `create` always

CountingSort/app_resources/prefix_sum_shader.comp.hlsl
```diff
@@ -0,0 +1,53 @@
...
+    {
+        bda::__ptr < uint32_t > (addr + sizeof(uint32_t) * index).template
+        deref().store(value);
+    }
+
+    bda::__spv_ptr_t<uint32_t> get_ptr(uint64_t index)
+    {
+        return bda::__ptr < uint32_t > (addr + sizeof(uint32_t) * index).template
+        deref().get_ptr();
+    }
```
devshgraphicsprogramming, line 32: `atomicAdd` instead to keep with accessors being storage space and underlying type agnostic

CountingSort/app_resources/scatter_shader.comp.hlsl
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    void set(uint64_t index, uint32_t value)
+    {
+        bda::__ptr < uint32_t > (out_addr + sizeof(uint32_t) * index).template
+        deref().store(value);
+    }
+
+    uint64_t in_addr, out_addr;
+};
```
devshgraphicsprogramming, line 60: if you made these templated I could let you slap them in a header like 
https://github.com/Devsh-Graphics-Programming/Nabla/blob/master/include/nbl/builtin/hlsl/memory_accessor.hlsl

and call them `bda_accessor`

but they';d need to:
1. be templated
2. guard the `atomicAdd` method with an `enable_if` checking for integral type and being 4 or 8 bytes in size

CountingSort/app_resources/prefix_sum_shader.comp.hlsl
devshgraphicsprogramming: lets move the conversation to the nabla PR, yes I've made a mistake here cause last invocation needs to send the sym to the first, my new pseudocode accounts for that

devshgraphicsprogramming:
looks nice, concise, short and dope

CountingSort/app_resources/prefix_sum_shader.comp.hlsl
```diff
@@ -0,0 +1,53 @@
+#include "nbl/builtin/hlsl/bda/__ptr.hlsl"
+#include "nbl/builtin/hlsl/sort/counting.hlsl"
+#include "nbl/builtin/hlsl/sort/common.hlsl"
+
+[[vk::push_constant]] nbl::hlsl::sort::CountingPushData pushData;
```
devshgraphicsprogramming, line 5: you need your own `common.hlsl` with your own push constant struct specifically for this example

CountingSort/app_resources/prefix_sum_shader.comp.hlsl
nipunG314: Done

nipunG314: Done

CountingSort/app_resources/prefix_sum_shader.comp.hlsl
nipunG314: Done

CountingSort/app_resources/prefix_sum_shader.comp.hlsl
nipunG314: Done

CountingSort/app_resources/scatter_shader.comp.hlsl
nipunG314: Done

CountingSort/app_resources/prefix_sum_shader.comp.hlsl
```diff
@@ -42,12 +39,20 @@ uint32_t3 nbl::hlsl::glsl::gl_WorkGroupSize()
...
-    PtrAccessor scratch_accessor = PtrAccessor::createAccessor(pushData.scratchAddress);
+    nbl::hlsl::sort::CountingParameters < uint32_t > params;
+    params.dataElementCount = pushData.dataElementCount;
+    params.elementsPerWT = pushData.elementsPerWT;
+    params.minimum = pushData.minimum;
+    params.maximum = pushData.maximum;
+
+    nbl::hlsl::sort::counting <WorkgroupSize, BucketCount, uint32_t, PtrAccessor, PtrAccessor, PtrAccessor, SharedAccessor> counter;
+    PtrAccessor input_accessor = PtrAccessor::create(pushData.inputKeyAddress);
+    PtrAccessor scratch_accessor = PtrAccessor::create(pushData.scratchAddress);
```
devshgraphicsprogramming, line 50: replace occurences of `scratch` in naming with `histogram`

devshgraphicsprogramming:
Also give the Example a Number, like 10

CountingSort/app_resources/prefix_sum_shader.comp.hlsl
nipunG314: Done

PR #107, opened 2024-04-18
Vulkan 1 3 example 03 fix

devshgraphicsprogramming:


03_DeviceSelectionAndSharedSources/app_resources/pplnLayoutMergeTest/shader_3.comp.hlsl
```diff
@@ -0,0 +1,15 @@
+#pragma wave shader_stage(compute)
+#include "common.hlsl"
+
+[[vk::binding(0,0)]] RWStructuredBuffer<SomeType> asdf;
+[[vk::binding(1,0)]] RWStructuredBuffer<SomeType> fdsa;
+[[vk::binding(2,0)]] RWByteAddressBuffer output;
```
devshgraphicsprogramming, line 6: `shader_1` and `shader_3` define this as a runtime descriptor array and 1 sized

Check if its legal to bind a descriptor with layout that has a Runtime Array binding flag to a SPIR-V shader that doesn't decorate the binding like that.

(in simpler words, can you use a descriptor layout with runtime array at binding 2 with `shader_3`)

03_DeviceSelectionAndSharedSources/main.cpp
```diff
@@ -13,318 +13,422 @@ using namespace system;
...
+	DeviceSelectionAndSharedSourcesApp(const path& _localInputCWD, const path& _localOutputCWD, const path& _sharedInputCWD, const path& _sharedOutputCWD) :
+		system::IApplicationFramework(_localInputCWD, _localOutputCWD, _sharedInputCWD, _sharedOutputCWD) {}
+
+	// we stuff all our work here because its a "single shot" app
+	bool onAppInitialized(smart_refctd_ptr<ISystem>&& system) override
+	{
+		// Remember to call the base class initialization!
+		if (!device_base_t::onAppInitialized(std::move(system)))
+			return false;
+		if (!asset_base_t::onAppInitialized(std::move(system)))
```
devshgraphicsprogramming, line 37: the `system` can be moved only once

03_DeviceSelectionAndSharedSources/main.cpp
```diff
@@ -13,318 +13,422 @@ using namespace system;
...
+			return false;
+
+		// TODO: run-time sized buffers are not supporten in hlsl.. do it when testing glsl
```
devshgraphicsprogramming, line 40: `ByteAddressBuffers` compile to equivalent of a runtime sized SSBO `GLSL` buffer with a single `uint32_t[]` block member

03_DeviceSelectionAndSharedSources/main.cpp
```diff
@@ -13,318 +13,422 @@ using namespace system;
...
 
-			// Nabla hardcodes the maximum descriptor set count to 4
-			constexpr uint32_t MaxDescriptorSets = 4; //SPhysicalDeviceLimits::MaxDescriptorSets;
-			smart_refctd_ptr<IGPUDescriptorSet> ds[MaxDescriptorSets];
+				// should merge successfully since shader is not messed up and it is the first merge
+			confirmExpectedOutput(pplnIntroData->merge(introspections[0].get()), true);
+				// should merge successfully since pipeline layout of "shader_2.comp.hlsl" is compatible with "shader_1.comp.hlsl"
+			confirmExpectedOutput(pplnIntroData->merge(introspections[1].get()), true);
+				// should not merge since pipeline layout of "shader_3.comp.hlsl" is not compatible with "shader_1.comp.hlsl"
+			confirmExpectedOutput(pplnIntroData->merge(introspections[2].get()), false);
```
devshgraphicsprogramming, line 97: because of the runtime Descriptor Array?

you already checked then?

03_DeviceSelectionAndSharedSources/app_resources/pplnLayoutMergeTest/shader_4.comp.hlsl
```diff
@@ -0,0 +1,14 @@
+#pragma wave shader_stage(compute)
+#include "common.hlsl"
+
+[[vk::binding(0,0)]] RWByteAddressBuffer output[];
+[[vk::binding(1,0)]] RWStructuredBuffer<SomeType> asdf;
```
devshgraphicsprogramming, line 5: make sure you leave comments in the shaders themselves if they're supposed to be broken, like here

03_DeviceSelectionAndSharedSources/main.cpp
```diff
@@ -13,318 +13,422 @@ using namespace system;
...
 			}
 
-			// a simple test to check we got the right thing back
-			auto outputData = mapBuffer(outputBuff,IDeviceMemoryAllocation::EMCAF_READ);
-			for (auto i=0; i<WorkgroupSize*WorkgroupCount; i++)
-			if (outputData[i]!=DWORDCount)
-				return logFail("DWORD at position %d doesn't match!\n",i);
-			m_device->unmapMemory(outputBuff->getBoundMemory());
-
-			return true;
```
devshgraphicsprogramming, line 318: lets not loose this and actually do some dispatches with one of the compute shaders to do actual work like we did originally

03_DeviceSelectionAndSharedSources/main.cpp
```diff
@@ -13,318 +13,422 @@ using namespace system;
...
+				{.dstSet = ds[1].get(),.binding = 2,.arrayElement = 0,.count = 2,.info = inputBuffersInfo},
+				{.dstSet = ds[3].get(),.binding = 6,.arrayElement = 0,.count = 1,.info = &outputBufferInfo}
 			};
-			
-			// We'll fill the buffers in a way they always add up to WorkgrouCount*WorkgroupSize
-			constexpr auto DWORDCount = WorkgroupSize*WorkgroupCount;
+
+			m_device->updateDescriptorSets(std::span(writes, 2), {});
+		}
+#endif
```
devshgraphicsprogramming, line 276: would be cool if you could use the Stage Introspection data to find out the values you need to populate `SWriteDescriptorSet` with (binding and set).

But I guess that would require you to "know" the names of the descriptor variables of the in the stage introspection.

03_DeviceSelectionAndSharedSources/main.cpp
```diff
@@ -13,318 +13,422 @@ using namespace system;
...
 
-			// Need to know input and output sizes by ourselves obviously
-			constexpr size_t WorkgroupCount = 4096;
-			constexpr size_t BufferSize = sizeof(uint32_t)*WorkgroupSize*WorkgroupCount;
+		// Nabla hardcodes the maximum descriptor set count to 4
+		constexpr uint32_t MaxDescriptorSets = ICPUPipelineLayout::DESCRIPTOR_SET_COUNT;
+		const std::array<IGPUDescriptorSetLayout*, MaxDescriptorSets> dscLayoutPtrs = { nullptr, dsLayout1.get(), nullptr, dsLayout3.get() };
+		std::array<smart_refctd_ptr<IGPUDescriptorSet>, MaxDescriptorSets> ds;
+		auto pool = m_device->createDescriptorPoolForDSLayouts(IDescriptorPool::ECF_NONE, std::span(dscLayoutPtrs.begin(), dscLayoutPtrs.end()));
+		pool->createDescriptorSets(dscLayoutPtrs.size(), dscLayoutPtrs.data(), ds.data());
```
devshgraphicsprogramming, line 190: should create from pipeline introspection

03_DeviceSelectionAndSharedSources/main.cpp
```diff
@@ -13,318 +13,422 @@ using namespace system;
...
+		{
+			IGPUComputePipeline::SCreationParams params = {};
+			params.layout = pplnLayout.get();
+			// Theoretically a blob of SPIR-V can contain multiple named entry points and one has to be chosen, in practice most compilers only support outputting one (and glslang used to require it be called "main")
+			params.shader.entryPoint = "main";
+			params.shader.shader = shader.get();
+			// we'll cover the specialization constant API in another example
+			if (!m_device->createComputePipelines(nullptr, { &params,1 }, &pipeline))
+				return logFail("Failed to create pipelines (compile & link shaders)!\n");
+		}
```
devshgraphicsprogramming, line 183: we should probably test-drive the instrospector here, and ignore
```
// Nabla actually has facilities for SPIR-V Reflection and "guessing" pipeline layouts for a given SPIR-V which we'll cover in a different example
```

the one thing you'll need to hand-roll here is creation of IGPUComputePipeline and IGPUDescriptorSet from ICPUComputePipeline and ICPUDescriptorSet along with all dependants (layouts and shader), without resorting to the asset converter (I wanna do that in example 09 or 24/25).

03_DeviceSelectionAndSharedSources/main.cpp
```diff
@@ -13,318 +13,422 @@ using namespace system;
...
-					cmdbuf->dispatch(WorkgroupCount,1,1);
-					cmdbuf->end();
-				}
+			{
+				auto* srcContent = spirvUnspecialized->getContent();
 
-				// A sidenote: Waiting for the Device or Queue to become idle does not give the same memory visibility guarantees as a signalled fence or semaphore
-				smart_refctd_ptr<IGPUFence> done = m_device->createFence(IGPUFence::ECF_UNSIGNALED);
+				system::ISystem::future_t<core::smart_refctd_ptr<system::IFile>> future;
+				m_physicalDevice->getSystem()->createFile(future, system::path("../app_resources/compiled.spv"), system::IFileBase::ECF_WRITE);
```
devshgraphicsprogramming, line 390: you probably want localCWD as the base path here

03_DeviceSelectionAndSharedSources/main.cpp
```diff
@@ -13,318 +13,422 @@ using namespace system;
...
+			// The nice thing is that when you load a shader from file, it has a correctly set `filePathHint`
+			// so it plays nicely with the preprocessor, and finds `#include`s without intervention.
+			options.preprocessorOptions.sourceIdentifier = source->getFilepathHint();
+			options.preprocessorOptions.logger = m_logger.get();
+			options.preprocessorOptions.includeFinder = compilerSet->getShaderCompiler(source->getContentType())->getDefaultIncludeFinder();
+
+			auto spirvUnspecialized = compilerSet->compileToSPIRV(source.get(), options);
+			const CSPIRVIntrospector::CStageIntrospectionData::SParams inspctParams = { .entryPoint = "main", .shader = spirvUnspecialized };
+
+			introspection = introspector.introspect(inspctParams);
```
devshgraphicsprogramming, line 379: do you debug print this anywhere?

Would be quite useful

03_DeviceSelectionAndSharedSources/main.cpp
```diff
@@ -13,318 +13,422 @@ using namespace system;
...
+					if (value != expectedValue)
+					{
+						m_logger->log("\"CSPIRVIntrospector::CPipelineIntrospectionData::merge\" function FAIL, incorrect output.",
+							ILogger::E_LOG_LEVEL::ELL_ERROR);
+					}
+					else
+					{
+						m_logger->log("\"CSPIRVIntrospector::CPipelineIntrospectionData::merge\" function SUCCESS, correct output.",
+							ILogger::E_LOG_LEVEL::ELL_PERFORMANCE);
+					}
```
devshgraphicsprogramming, line 76: imo you should do the merge of stage to pipeline in the lambda and debug print somewhere the pipeline layout introspection post-merge

03_DeviceSelectionAndSharedSources/main.cpp
devshgraphicsprogramming: ah you don't, please use builtin resources CMake, then you can use "app_resources/" cause app_resources will be a mountpoint

03_DeviceSelectionAndSharedSources/app_resources/pplnLayoutMergeTest/shader_4.comp.hlsl
devshgraphicsprogramming: this shader won't even be compatible with itself (merge with a blank pipeline intro should fail)

03_DeviceSelectionAndSharedSources/main.cpp
Przemog1: i can only create CPU descriptor set layouts from introspection, is there any way to cast from `ICPUDescriptorSetLayout` to `IGPUDescriptorSetLayout`?

devshgraphicsprogramming: no you can't cast, you need to create by hand.

IF you get the example to a mergable state, I can use the asset-converter here maybe.

03_DeviceSelectionAndSharedSources/main.cpp
Przemog1: that will require additional `debugPrint` function in `CPipelineIntrospectionData`

PR #108, opened 2024-04-22
Config profiles mapping

devshgraphicsprogramming:


devshgraphicsprogramming:
Could be a nice base for @Fletterio's Shader Cache

PR #112, opened 2024-05-03
Changed examples to follow separable sampler changes

Fletterio:
Slightly modified example 07 to test the sampled image in action


devshgraphicsprogramming:
you can try using separable sampler+image in example 24

PR #113, opened 2024-05-04
Example 14 ComputeScan fix for new NBL API and use HLSL

kpentaris:
Still doesn't work. There are shader issues with shared memory reading. The basic outline is correct however.


14_ComputeReduce/main.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+		m_logger->log("Input element count: %d", ILogger::ELL_PERFORMANCE, in_count);
+
+		inputData = new uint32_t[in_count];
+		{
+			std::random_device random_device;
+			std::mt19937 generator(random_device());
+			std::uniform_int_distribution<uint32_t> distribution(0u, ~0u);
+			for (auto i = 0u; i < in_count; i++)
+				inputData[i] = distribution(generator) % 100000;
```
devshgraphicsprogramming, line 43: you might want to make these small enough so the reduction or scan doesn't overflow (I think our algo is not necessarily resillient to that)

14_ComputeReduce/main.cpp
```diff
@@ -0,0 +1,281 @@
...
+		}
+		auto minSSBOAlign = m_physicalDevice->getLimits().minSSBOAlignment;
+		constexpr auto begin = in_count / 4 + 118;
+		assert(((begin * sizeof(uint32_t)) & (minSSBOAlign - 1u)) == 0u);
+		constexpr auto end = in_count * 3 / 4 - 78;
+		assert(((end * sizeof(uint32_t)) & (minSSBOAlign - 1u)) == 0u);
+		constexpr auto elementCount = end - begin;
```
devshgraphicsprogramming, line 50: what's the point of all this?

14_ComputeReduce/main.cpp
```diff
@@ -0,0 +1,281 @@
...
+			.semaphore = semaphore.get(),
+			.value = 1,
+			.stageMask = PIPELINE_STAGE_FLAGS::ALL_TRANSFER_BITS
+		} };
+
+		smart_refctd_ptr<IGPUBuffer> gpuinputDataBuffer;
+		{
+			IGPUBuffer::SCreationParams inputDataBufferCreationParams = {};
+			inputDataBufferCreationParams.size = sizeof(uint32_t) * in_count; // TODO Declare the element data type in the shader?
+			inputDataBufferCreationParams.usage = IGPUBuffer::EUF_STORAGE_BUFFER_BIT | IGPUBuffer::EUF_TRANSFER_SRC_BIT | IGPUBuffer::EUF_TRANSFER_DST_BIT;
```
devshgraphicsprogramming, line 64: address usage, if you want to use BDA

14_ComputeReduce/main.cpp
```diff
@@ -0,0 +1,281 @@
...
+			}
+			gpuinputDataBuffer = *temp.get();
+		}
+		SBufferRange<IGPUBuffer> in_gpu_range = { begin * sizeof(uint32_t), elementCount * sizeof(uint32_t), gpuinputDataBuffer };
+
+		auto reducer = m_utils->getDefaultReducer();
+
+		CArithmeticOps::DefaultPushConstants reduce_push_constants;
+		CArithmeticOps::DispatchInfo reduce_dispatch_info;
+		reducer->buildParameters(elementCount, reduce_push_constants, reduce_dispatch_info);
```
devshgraphicsprogramming, line 88: I think I've left a comment in the "mother PR" that this should be a constructor of parameters or something

Kinda silly to have to use a factory to do some basic arithmetic

14_ComputeReduce/main.cpp
```diff
@@ -0,0 +1,281 @@
...
+		
+		{
+			IGPUBuffer::SCreationParams params = {};
+			params.size = in_gpu_range.size;
+			params.usage = IGPUBuffer::EUF_TRANSFER_DST_BIT;
+			// (REVIEW): Check if this new download_buffer is needed or if we can directly read from the gpu_input buffer
+			auto downloaded_buffer = m_device->createBuffer(std::move(params));
+			auto memReqs = downloaded_buffer->getMemoryReqs();
+			memReqs.memoryTypeBits &= m_physicalDevice->getDownStreamingMemoryTypeBits();
+			auto queriesMem = m_device->allocate(memReqs, downloaded_buffer.get());
```
devshgraphicsprogramming, line 169: use the downstreaming buffer and make a consumption callback.

Otherwise you'll have a problem making a 128MB HOST_VISIBLE and DEVICE_LOCAL buffer, not everyone has ReBAR configured correctly.

devshgraphicsprogramming: See example 05 (whichever does the streaming consumer callback), how to download a buffer and consume it "piecewise"

PR #114, opened 2024-05-15
CAD emulated float64

Przemog1:
https://github.com/Devsh-Graphics-Programming/Nabla/pull/696


22_CppCompat/app_resources/emulated_float64_t_test/common.hlsl
```diff
@@ -0,0 +1,49 @@
...
+
+#include <nbl/builtin/hlsl/cpp_compat.hlsl>
+#include <nbl/builtin/hlsl/emulated_float64_t.hlsl>
+
+NBL_CONSTEXPR uint32_t WORKGROUP_SIZE = 256;
+
+emulated::emulated_float64_t _static_cast(const float val)
+{
+    emulated::emulated_float64_t retval;
+    retval.data = val; // TODO: manually upgrade the `val` IEEE754 32bit pattern to 64bit
```
devshgraphicsprogramming, line 13: job of the `create` static method, here you just forward to it

22_CppCompat/main.cpp
```diff
@@ -366,6 +416,358 @@ class CompatibilityTest final : public MonoDeviceApplication, public MonoAssetMa
...
+            .lessVal = false,
+            .greaterVal = true,
+            .convertionToBoolVal = true,
+            .convertionToIntVal = 20,
+            .convertionToUint32Val = 20u,
+            .convertionToUint64Val = 20ull,
+            .convertionToFloatVal = 20.0f,
+            .convertionToDoubleVal = 20.0,
+            //.convertionToHalfVal = 20;
+        };
```
devshgraphicsprogramming, line 607: what you could do is use xoroshiro64** or somethign similar to come up with 64bit patterns denoting individual lhs and rhs values, then runs same code on CPU and GPU for a few hundred/thousand "random" doubles and see if you match

Obviously the `create` with different data types, and handling of Inf and NaN in operators needs to be tested specially, but you get the flow.

The more of the operation/input matching code you can hoist to a common HLSL/C++ header the better.

62_CAD/common.hlsl
```diff
@@ -41,14 +42,14 @@ struct DrawObject
 
 struct LinePointInfo
 {
-    float64_t2 p;
+    nbl::hlsl::portable_vector64_t2 p;
```
Erfan-Ahmadi, line 45: a bit ugly but I think this is necessary cause other choice would be for this to stay a float64_t2 and then in vtx shader construct a portable_vector64_t2 when loading these values and I assume constructing one is expensive (when float64 feature is disabled)

62_CAD/common.hlsl
```diff
@@ -41,14 +42,14 @@ struct DrawObject
...
     float32_t phaseShift;
     float32_t stretchValue;
 };
 
 struct QuadraticBezierInfo
 {
-    nbl::hlsl::shapes::QuadraticBezier<float64_t> shape; // 48bytes = 3 (control points) x 16 (float64_t2)
+    nbl::hlsl::shapes::QuadraticBezier<nbl::hlsl::portable_float64_t<> > shape; // 48bytes = 3 (control points) x 16 (emulated_float64_t)
```
Erfan-Ahmadi, line 52: what's with the extra <>, template stuff compile error?

62_CAD/common.hlsl
devshgraphicsprogramming: you cannot have any `float64_t` (except CONSTANT LITERALS or anything else that gets constant folded into not-a-float64) if you want to avoid emitting the `shaderFloat64` capability in the SPIR-V

62_CAD/common.hlsl
devshgraphicsprogramming: room for device capability traits (the template param gets defaulted to `void`), the JIT device traits should be passed here once all the emulated float64 testing is complete

62_CAD/vertex_shader.hlsl
```diff
@@ -28,9 +28,9 @@ ClipProjectionData getClipProjectionData(in MainObject mainObj)
     if (mainObj.clipProjectionAddress != InvalidClipProjectionAddress)
     {
         ClipProjectionData ret;
-        ret.projectionToNDC = vk::RawBufferLoad<float64_t3x3>(mainObj.clipProjectionAddress, 8u);
-        ret.minClipNDC      = vk::RawBufferLoad<float32_t2>(mainObj.clipProjectionAddress + sizeof(float64_t3x3), 8u);
-        ret.maxClipNDC      = vk::RawBufferLoad<float32_t2>(mainObj.clipProjectionAddress + sizeof(float64_t3x3) + sizeof(float32_t2), 8u);
+        ret.projectionToNDC = vk::RawBufferLoad<nbl::hlsl::portable_matrix64_t3x3>(mainObj.clipProjectionAddress, 8u);
+        ret.minClipNDC      = vk::RawBufferLoad<float32_t2>(mainObj.clipProjectionAddress + sizeof(nbl::hlsl::portable_matrix64_t3x3), 8u);
+        ret.maxClipNDC      = vk::RawBufferLoad<float32_t2>(mainObj.clipProjectionAddress + sizeof(nbl::hlsl::portable_matrix64_t3x3) + sizeof(float32_t2), 8u);
```
devshgraphicsprogramming, line 33: this example would really benefit from `using namespace nbl::hlsl` in the `common.hlsl` to reduce explicit namespace clutter

22_CppCompat/app_resources/emulated_float64_t_test/common.hlsl
```diff
@@ -0,0 +1,64 @@
...
+    typename nbl::hlsl::emulated_float64_t<FastMath, FlushDenormToZero>::storage_t substractionVal;
+    typename nbl::hlsl::emulated_float64_t<FastMath, FlushDenormToZero>::storage_t multiplicationVal;
+    typename nbl::hlsl::emulated_float64_t<FastMath, FlushDenormToZero>::storage_t divisionVal;
+
+    // relational
+    int lessOrEqualVal;
+    int greaterOrEqualVal;
+    int equalVal;
+    int notEqualVal;
+    int lessVal;
```
devshgraphicsprogramming, line 62: ping me when ex 22 contains either no changes or only improvements that have nothing to do with emulated float

62_CAD/common.hlsl
Przemog1: i already do emulated_float64_t<nbl::hlsl::jit::device_capabilities>

devshgraphicsprogramming: do not introduce a dependency on `jit` namespace outside the examples

22_CppCompat/app_resources/emulated_float64_t_test/common.hlsl
Przemog1: yeah, forgot to delete some stuff.. i will make this file exactly the same as it is on master

Przemog1: except for one bug i found

62_CAD/common.hlsl
Przemog1: so far only vertex shader of example 62 is dependent on jit

devshgraphicsprogramming: @Erfan-Ahmadi resolve when read

62_CAD/common.hlsl
devshgraphicsprogramming: mostly a reply to @Erfan-Ahmadi , but thanks for the emote @Przemog1

62_CAD/vertex_shader.hlsl
```diff
@@ -7,6 +7,11 @@
 #include <nbl/builtin/hlsl/algorithm.hlsl>
 #include <nbl/builtin/hlsl/jit/device_capabilities.hlsl>
 
+using ActualFloat64 = portable_float64_t<jit::device_capabilities>;
```
Erfan-Ahmadi, line 10: can you add a mode/macro "ForceEmulatedFloat64" for us to test on GPUs that already have Float64 feature?
or alternatively tell me how you usually test emulated vs non-emulated rapidly?

62_CAD/main.cpp
```diff
@@ -68,7 +68,7 @@ constexpr std::array<float, (uint32_t)ExampleMode::CASE_COUNT> cameraExtents =
 	600.0,	// CASE_8
 };
 
-constexpr ExampleMode mode = ExampleMode::CASE_7;
```
Erfan-Ahmadi, line 71: revert these changes please, I changed shader compilation a lot, it will introduce conflicts

62_CAD/common.hlsl
```diff
@@ -1,11 +1,30 @@
...
 #ifdef __HLSL_VERSION
 #include <nbl/builtin/hlsl/math/equations/quadratic.hlsl>
+#include <nbl/builtin/hlsl/jit/device_capabilities.hlsl>
+#endif
+
+using namespace nbl::hlsl;
+
+// because we can't use jit/device_capabilities.hlsl in c++ code
+#ifdef __HLSL_VERSION
+using cpp_hlsl_portable_float64_t = portable_float64_t<jit::device_capabilities>;
```
Erfan-Ahmadi, line 19: hmm, you're redoing this again with different names in vertex_shader
choose a single name for it and use the same in VertexShader.
`PortableFloat64` instead of `ActualFloat64` seems like a good choice. also the Pascal Case naming indicates it might actually be a struct/class emulating it

62_CAD/vertex_shader.hlsl
Przemog1: https://github.com/Devsh-Graphics-Programming/Nabla/blob/emulated_float64_t/include/nbl/builtin/hlsl/portable/float64_t.hlsl#L13
```
// force emulated_float64<true, true>
using portable_float64_t = typename conditional<device_capabilities_traits<device_caps>::shaderFloat64, emulated_float64_t<true, true>, emulated_float64_t<true, true> >::type;

// force emulated_float64<true, true>
using portable_float64_t = typename conditional<device_capabilities_traits<device_caps>::shaderFloat64, float64_t, emulated_float64_t<true, true> >::type;
```

I actually forgot to change that back, so emulated float is forced

Erfan-Ahmadi: then I think you can do this `ForceEmulatedFloat64` option by using a fake caps struct

03_DeviceSelectionAndSharedSources/app_resources/float64_t_test.hlsl
```diff
@@ -0,0 +1,15 @@
...
+#include "common.hlsl"
+#include <nbl/builtin/hlsl/emulated_float64_t.hlsl>
+
+[numthreads(256,1,1)]
+void main() 
+{
+    const emulated::float64_t a = _static_cast<emulated::float64_t>(6.9f);
+    const emulated::float64_t b = _static_cast<emulated::float64_t>(4.5f);
+
+    vk::RawBufferStore<emulated::float64_t::storage_t>(0,(a*b).data);
```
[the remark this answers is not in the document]
devshgraphicsprogramming: it kinda looks like a file someone forgot to remove (no chanes in 03 main.cpp)

62_CAD/shaders/main_pipeline/vertex_shader.hlsl
```diff
@@ -39,14 +40,18 @@ ClipProjectionData getClipProjectionData(in MainObject mainObj)
...
 
-float32_t2 transformPointScreenSpace(float64_t3x3 transformation, uint32_t2 resolution, float64_t2 point2d) 
+// TODO[Przemok]: Fix/Remove your todo here?!
+float2 transformPointScreenSpace(pfloat64_t3x3 transformation, uint32_t2 resolution, pfloat64_t2 point2d)
 {
-    float64_t2 ndc = transformPointNdc(transformation, point2d);
-    return (float32_t2)((ndc + 1.0) * 0.5 * resolution);
+    pfloat64_t2 ndc = transformPointNdc(transformation, point2d);
+    // TODO: "ndc + 1" doesn't work, probably because of `emulated_float64_t::create(float val)`
+    pfloat64_t2 result = (ndc + 1.0f) * 0.5f * _static_cast<pfloat64_t2>(resolution);
```
devshgraphicsprogramming, line 48: + int shouldn't work, I don't like it

and still because HLSL doesn't do conversion operators you're unable to get an implicit conversion to satisfy an argument type.

In C++ `double+int` compiles because `double` has an implicit constructor from `int` and only `operator+(const double)`

NOT because an overload of `+` exists for every other type (float, int, etc.) as that would wreck mayhem everywhere else (every function that expects a `double` would throw a fit if you try to feed an integer into it.

What I want to avoid is `double+int` working and then people asking why `f(double)` doesn't work when you call with `f(int)`

64_EmulatedFloatTest/app_resources/benchmark/benchmark.comp.hlsl
```diff
@@ -0,0 +1,131 @@
...
+// int from 0 to 69 = 3.11133�10^30
+
+template<typename F64>
+struct Random16thPolynomial
+{
+	void randomizeCoefficients()
+	{
+		Xoroshiro64Star rng = Xoroshiro64Star::construct(69);
+
+		// can't just do `coefficients[i] = rng()` this will create retarded numbers or special value exponents
```
devshgraphicsprogramming, line 31: language and innuendo son.

devshgraphicsprogramming: P.S. I'm sure the numbers would be "fine" if you wiped the position 62 and 61 bits to 0

devshgraphicsprogramming:
Has my seal of approval

62_CAD/vertex_shader.hlsl
Przemog1:
```cpp
#ifdef NBL_FORCE_EMULATED_FLOAT_64
using portable_float64_t = emulated_float64_t<true, true>;
#else
using portable_float64_t = typename conditional<device_capabilities_traits<device_caps>::shaderFloat64, float64_t, emulated_float64_t<true, true> >::type;
#endif
```
did this, so in order to force ef64 one has to define `NBL_FORCE_EMULATED_FLOAT_64` macro before including `portable/float64.hlsl`

PR #116, opened 2024-05-31
Hatch fill patterns for the CAD example

deprilula28:
TODO list:

- [ ] Vulkan has a validation error that I haven't really made sense of
- [ ] Need to decide how MsdfFillPattern will work (is it going to be fixed like I made it?)


62_CAD/DrawResourcesFiller.cpp
```diff
@@ -435,66 +440,97 @@ void DrawResourcesFiller::finalizeTextureCopies(SIntendedSubmitInfo& intendedNex
...
+
+	if (!textureCopies.size())
+		return;
+
 	// preparing images for copy
-	// @Lucas TODO: add a vector<image_barrier_t> push_back a barrier like below for each of the indices in `textureCopies`
-	IGPUCommandBuffer::SPipelineBarrierDependencyInfo::image_barrier_t barriersBeforeCopy[] =
+	using image_barrier_t = IGPUCommandBuffer::SPipelineBarrierDependencyInfo::image_barrier_t;
+	std::vector<image_barrier_t> barriers;
+	barriers.reserve(textureCopies.size());
```
Erfan-Ahmadi, line 451: remove this reserve textureCopies.size()?
you decided to do all in 1 barrier, that's ok.

Erfan-Ahmadi: updated comment

62_CAD/DrawResourcesFiller.h
```diff
@@ -24,6 +25,24 @@ struct DrawBuffers
 	smart_refctd_ptr<BufferType> lineStylesBuffer;
 };
 
+// TODO: Better place for this
+enum class MsdfFillPattern: uint32_t
```
Erfan-Ahmadi, line 29: This is more high-level than DrawResourcesFiller, so make that and the factory functions that give you msdfs into a seperate file. this way we can easily copy paste into n4ce and use that there.

62_CAD/main.cpp
```diff
@@ -457,6 +525,9 @@ class ComputerAidedDesign final : public examples::SimpleWindowedApplication, pu
 	inline bool onAppInitialized(smart_refctd_ptr<ISystem>&& system) override
 	{
 		m_inputSystem = make_smart_refctd_ptr<InputSystem>(logger_opt_smart_ptr(smart_refctd_ptr(m_logger)));
+			LRUCache<int, char> hugeCache(50000000u);
+			hugeCache.insert(0, '0');
+
```
Erfan-Ahmadi, line 530: old test code?

62_CAD/DrawResourcesFiller.h
Erfan-Ahmadi: We'll have classes/structs related to text placement and msdf generations anyways, it makes sense to put them all in there along with the hashing of the data

deprilula28: When doing it one by one, the validation layer kept claiming that the resource was on the wrong layout, we could make it transition from shader read, and only use undefined for the first transition

Erfan-Ahmadi: UNDEFINED is not a wrong layout.
Your next barrier after the upload was the issue. Because some of the textures were freshly TRANSFER_DST and some of them were old SHADER_READONLY
Doing both the before and after barriers for only the uploading ones shouldn't result in a validation error. 
Any transition barrier chain which starts from UNDEFINED, should be pretty straightforward because you're completely overwriting the previous slot

62_CAD/main.cpp
```diff
@@ -350,6 +351,73 @@ class ComputerAidedDesign final : public examples::SimpleWindowedApplication, pu
...
+		msdfgen::generateMTSDF(msdfMap, glyph, MsdfPixelRange, { scaleX, scaleY }, msdfgen::Vector2(0.0, 0.0));
+
+		// TODO: Optimize this
+		auto cpuBuf = core::make_smart_refctd_ptr<ICPUBuffer>(glyphW * glyphH * 4);
+		uint8_t* data = reinterpret_cast<uint8_t*>(cpuBuf->getPointer());
+		for (int y = 0; y < msdfMap.height(); ++y)
+		{
+			for (int x = 0; x < msdfMap.width(); ++x)
+			{
+				auto pixel = msdfMap(x, glyphH - 1 - y);
```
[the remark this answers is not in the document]
Erfan-Ahmadi: Ok my bad, brainfart! it should be filled with uint8_t

BUTTTT you can do format conversions WHILE uploading with our image upload utility. (change srcFormat to UINT in the input params)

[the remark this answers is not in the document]
Erfan-Ahmadi: @devshgraphicsprogramming can you verify this? I forgot the vk format stuff regarding conversions.

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -280,44 +279,50 @@ void DrawResourcesFiller::drawHatch(const Hatch& hatch, const float32_t4& color,
 	drawHatch(hatch, color, InvalidTextureHash, intendedNextSubmit);
 }
 
-void DrawResourcesFiller::addMSDFTexture(ICPUBuffer const* srcBuffer, const asset::IImage::SBufferCopy& region, texture_hash hash, SIntendedSubmitInfo& intendedNextSubmit)
+void DrawResourcesFiller::addMSDFTexture(ICPUBuffer const* srcBuffer, uint64_t bufferOffset, uint32_t3 imageExtent, texture_hash hash, SIntendedSubmitInfo& intendedNextSubmit)
```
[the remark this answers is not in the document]
deprilula28: Done

62_CAD/DrawResourcesFiller.h
deprilula28: Done (I think commented on the wrong place here)

62_CAD/main.cpp
deprilula28: Removed

62_CAD/MSDFs.h
```diff
@@ -0,0 +1,64 @@
...
+		{
+		case MsdfTextureType::HATCH_FILL_PATTERN:
+			textureHash = std::hash<uint32_t>{}(uint32_t(s.fillPattern));
+			break;
+		case MsdfTextureType::FONT_GLYPH:
+			textureHash = std::hash<uint32_t>{}(s.glyphIndex);
+			break;
+		}
+
+		return textureTypeHash ^ (textureHash << 1);
```
Erfan-Ahmadi, line 57: use `core::hash_combine`

62_CAD/MSDFs.h
```diff
@@ -0,0 +1,64 @@
...
+		std::size_t textureTypeHash = std::hash<uint32_t>{}(uint32_t(s.textureType));
+		std::size_t textureHash;
+
+		switch (s.textureType) 
+		{
+		case MsdfTextureType::HATCH_FILL_PATTERN:
+			textureHash = std::hash<uint32_t>{}(uint32_t(s.fillPattern));
+			break;
+		case MsdfTextureType::FONT_GLYPH:
+			textureHash = std::hash<uint32_t>{}(s.glyphIndex);
```
Erfan-Ahmadi, line 53: then this path becomes textureHash = s.glyphHash

62_CAD/MSDFs.cpp
```diff
@@ -0,0 +1,464 @@
...
+	auto shapeBounds = glyph.getBounds();
+
+	msdfgen::edgeColoringSimple(glyph, 3.0); // TODO figure out what this is
+	msdfgen::Bitmap<float, 4> msdfMap(glyphW, glyphH);
+
+	float scaleX = (1.0 / float(glyphExtents.x)) * glyphW;
+	float scaleY = (1.0 / float(glyphExtents.y)) * glyphH;
+	msdfgen::generateMTSDF(msdfMap, glyph, MsdfPixelRange, { scaleX, scaleY }, msdfgen::Vector2(0.0, 0.0));
+
+	// TODO: Optimize this
```
Erfan-Ahmadi, line 31: as per discussions in discord, you can remove the for loops and set srcFormat in image upload utility to RGBA8_SFLOAT to get the conversion while memcpying to staging memory with our image filters (all happening inside the util)

62_CAD/common.hlsl
```diff
@@ -197,6 +197,7 @@ NBL_CONSTEXPR uint32_t InvalidTextureIdx = nbl::hlsl::numeric_limits<uint32_t>::
 NBL_CONSTEXPR MajorAxis SelectedMajorAxis = MajorAxis::MAJOR_Y;
 // TODO: get automatic version working on HLSL
 NBL_CONSTEXPR MajorAxis SelectedMinorAxis = MajorAxis::MAJOR_X; //(MajorAxis) (1 - (uint32_t) SelectedMajorAxis);
+NBL_CONSTEXPR float MsdfPixelRange = 4.0;
```
Erfan-Ahmadi, line 200: What is MsdfPixelRange?

62_CAD/MSDFs.h
deprilula28: Done

62_CAD/MSDFs.h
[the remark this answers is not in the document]
deprilula28: Done

deprilula28: Done

62_CAD/common.hlsl
deprilula28: This is how many pixels, in MSDF resolution (32x32), of distance away (or inside) the shape that can be represented as a signed distance

(From msdfgen docs: specifies the width of the range around the shape between the minimum and maximum representable signed distance in shape units or distance field pixels, respectivelly.)

62_CAD/fragment_shader.hlsl
```diff
@@ -372,6 +372,26 @@ float32_t4 calculateFinalColor<true>(const uint2 fragCoord, const float localAlp
     return col;
 }
 
+// TODO: place this MSDF code elsewhere
+float median(float r, float g, float b) {
+    return max(min(r, g), min(max(r, g), b));
+}
+
+float screenPxRange(float2 uv) {
```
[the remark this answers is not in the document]
deprilula28: It's basically this:

`On screen size of the quad` / `MSDF Resolution` * `MSDF Pixel Range`

(From msdfgen docs: `screenPxRange()` represents the distance field range in output screen pixels. For example, if the pixel range was set to 2 when generating a 32x32 distance field, and it is used to draw a quad that is 72x72 pixels on the screen, it should return 4.5 (because 72/32 * 2 = 4.5). For 2D rendering, this can generally be replaced by a precomputed uniform value.)

I did it using the 3D path as we have to support transformations on the text (due to DWG and such), so this allows it to be more flexible (and only costs one fwidth)

62_CAD/MSDFs.cpp
deprilula28: Done

62_CAD/main.cpp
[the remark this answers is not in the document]
deprilula28: Done

62_CAD/DrawResourcesFiller.h
```diff
@@ -55,13 +56,21 @@ struct DrawResourcesFiller
...
 	// ! return index to be used later in hatch fill style or text glyph object
-	void addMSDFTexture(ICPUBuffer const* srcBuffer, const asset::IImage::SBufferCopy& region, texture_hash hash, SIntendedSubmitInfo& intendedNextSubmit);
+	struct MsdfTextureUploadInfo 
+	{
+		core::smart_refctd_ptr<ICPUBuffer> cpuBuffer;
+		uint64_t bufferOffset;
+		uint32_t3 imageExtent;
+		std::vector<CPolyline> polylines;
+	};
+	void addMSDFTexture(std::function<MsdfTextureUploadInfo()> createResourceIfEmpty, texture_hash hash, SIntendedSubmitInfo& intendedNextSubmit);
```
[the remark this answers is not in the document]
deprilula28: Done

62_CAD/DrawResourcesFiller.cpp
deprilula28: Done

62_CAD/common.hlsl
Erfan-Ahmadi: oh okay I got it, because we're using a [0,1] range we need to map it somehow to [-MSDFRange/2, +MSDFRange/2]
and that gets you the distance in "texels" if you're inside the 32x32 image

62_CAD/fragment_shader.hlsl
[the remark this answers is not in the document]
Erfan-Ahmadi: I think it should be / float2(32, 32); or the msdf texture size instead of `float2(globals.resolution);`
let's double check the math, so we're trying to get from

1. from [0, 1] of msdf sampled value to [-0.5, 0.5] by `(sd - 0.5)`

2. then we multiply by `MsdfPixelRange` to get it to [-MsdfPixelRange/2, MsdfPixelRange/2] and that is basically the 
distance in glyph texture units (because the spectrum/range from lowest to highest sdf is MsdfPixelRange! the input to msdfgen!)
now we need this distance in uv units:

3. we divde by (32,32) or `GetDimensions(msdfTexture)` to get the distance in uv units
now how to turn uv distance into actual screenspace distance?! that's the main question!

here it uses fwidth to do that because it shows you the rate of change of both U and V with regards to x and y on the screen and the inverse of that gives you the screenspace size of the
so if everything is aligned (u with x and v with y) it becomes `1.0 / abs(float2(ddx(u),ddy(v)))`

`1/ddx(u)` gives you the horizontal size of the glyph box in screen pixels!
and `1/ddy(v)` gives you the vertical size of the glyph box in screen pixels!
If you're not sure why read the thought experiment below:

++++
Optional Read 
ddx(u) tells you the rate of change of u per pixel in x direction
if you think about it ddx(u) is basicaly `dU/dScreenPixel` (if you go 1 pixel how much does u value change?)
and 1/ddx(u) is `dScreenPixel/dU` and gives you the inverse rate of change (if you go by 1 units in `u` how many units in screen pixels? ---> width pixels!) 
++++

But fwidth is just used as an approximation. `float2 fwidth = abs(ddx(uv) + ddy(uv))`
small note: ddx(uv) is a float2, it's computing ddx for both u and v.

Now that you understand this, you get why is the `/ float2(globals.resolution)` is buggy and why you need to have different functions for this for 1. glyph boxes 2. hatch fills
and if you fix this you are going to see these bugs fixed:
1. some hatch fills msdfs being buggy or non connected 
2. text being too blurry when zooming in 

A. How to handle this function for "Hatch Box Fills"?
for hatch box fills we cannot use this `fwidth(uv)` because we're doing repeat sampling of a 8x8 msdf onto the screen! (that's ptobably you caused a double bug by the wrong division, two wrongs sometimes make a right :D)
so for hatch fills it's `float2 screenTexSize = float2(8,8)` or whatever the fixed size for them is!

B. How to handle this function for "glyph boxes"?
for glyph boxes (or I think you call the fontglyphs in other branch), the screenspace size of the glyph box can be computed in vshader and get passed along! no need for fwidth, and it will result in a nice shape with any rotation because the sdf would be exact as the last step (3rd one) above get's you correctly from uv distance to screenspace distance, then you can use that to compute opacity or whatever

62_CAD/MSDFs.cpp
```diff
@@ -0,0 +1,452 @@
...
+	}
+
+	auto shapeBounds = glyph.getBounds();
+
+	msdfgen::edgeColoringSimple(glyph, 3.0); // TODO figure out what this is
+	msdfgen::Bitmap<float, 4> msdfMap(glyphW, glyphH);
+
+	float scaleX = (1.0 / float(glyphExtents.x)) * glyphW;
+	float scaleY = (1.0 / float(glyphExtents.y)) * glyphH;
+	msdfgen::generateMTSDF(msdfMap, glyph, MsdfPixelRange, { scaleX, scaleY }, msdfgen::Vector2(0.0, 0.0));
```
Erfan-Ahmadi, line 29: why are you generating MTSDF (it has true sdf in alpha channel) but don't use it? you can just make the MSDF textures RGB and generateMSDF instead of MTSDF

deprilula28: It was for debugging, easier to know what's wrong with the texture in render doc that way, but I'll change it over when everything is working

deprilula28: Msdfs are harder to understand and read

Erfan-Ahmadi: Or you can just use msdf.exe, generate the msdf and compare

deprilula28: I can't use that with the hatch fills, since they're not exactly font files, but I'll change it over to msdf only when everything works

62_CAD/vertex_shader.hlsl
```diff
@@ -455,6 +455,30 @@ PSInput main(uint vertexID : SV_VertexID)
...
+        const float2 screenDirU = (float2) transformVectorNdc(clipProjectionData.projectionToNDC, dirU);
+        const float2 screenDirV = (float2) transformVectorNdc(clipProjectionData.projectionToNDC, dirV);
+
+        float2 corner = float2(bool2(vertexIdx & 0x1u, vertexIdx >> 1));
+        const float2 coord = screenTopLeft + corner * (screenDirU + screenDirV);
+
+        // TODO needs to handle rotations as well, probably from curve box code
+        const float2 ndcAxisMin = screenTopLeft;
+        const float2 ndcAxisMax = screenTopLeft + screenDirU + screenDirV;
+        const float2 screenSpaceAabbExtents = abs(ndcAxisMax - ndcAxisMin) * float2(globals.resolution);
```
Erfan-Ahmadi, line 475: Just use lenght(screenDirU) and length(screenDirV) , to get the length of the sides

62_CAD/vertex_shader.hlsl
```diff
@@ -455,6 +455,30 @@ PSInput main(uint vertexID : SV_VertexID)
...
+        float64_t2 dirU = vk::RawBufferLoad<double2>(drawObj.geometryAddress + sizeof(double2), 8u);
+        float64_t2 dirV = vk::RawBufferLoad<double2>(drawObj.geometryAddress + sizeof(double2) * 2, 8u);
+        uint32_t textureId = vk::RawBufferLoad<uint32_t>(drawObj.geometryAddress + sizeof(double2) * 3, 4u);
+
+        const float2 screenTopLeft = (float2) transformPointNdc(clipProjectionData.projectionToNDC, topLeft);
+        const float2 screenDirU = (float2) transformVectorNdc(clipProjectionData.projectionToNDC, dirU);
+        const float2 screenDirV = (float2) transformVectorNdc(clipProjectionData.projectionToNDC, dirV);
+
+        float2 corner = float2(bool2(vertexIdx & 0x1u, vertexIdx >> 1));
+        const float2 coord = screenTopLeft + corner * (screenDirU + screenDirV);
```
Erfan-Ahmadi, line 470: This should be topLeft + corner.x * dirU + corner.y * dirV
Now rotation is handled

62_CAD/vertex_shader.hlsl
Erfan-Ahmadi: Also, not aabb anymore. Use box or obb

62_CAD/fragment_shader.hlsl
```diff
@@ -372,6 +372,23 @@ float32_t4 calculateFinalColor<true>(const uint2 fragCoord, const float localAlp
     return col;
 }
 
+// TODO: place this MSDF code elsewhere
```
Erfan-Ahmadi, line 375: These can be in nbl/hlsl/text_rendering/msdf.hlsl within correct namespace.
(Fix indentation please with the brackets) 

Do this when you've finalised working on these functions cause otherwise built-ins have to rebuild each time you make a change to the hlsl built-in.

Don't forget to document what the different input params and terms such as msdfPixelRange means and how it's related to the msdfgen lib (you can copy paste and quote from the lib above these functions)

62_CAD/fragment_shader.hlsl
```diff
@@ -372,6 +372,23 @@ float32_t4 calculateFinalColor<true>(const uint2 fragCoord, const float localAlp
...
+float median(float r, float g, float b) {
+    return max(min(r, g), min(max(r, g), b));
+}
+
+float screenPxRange(float2 screenSpaceSizePixels, float2 distanceFieldSizePixels) {
```
Erfan-Ahmadi, line 380: Let's not bring this function to the builtin hlsl header, because it's application dependant. Let's be flexible like msdfgen about user providing a screenPxRange. But remember to document good there so there is no confusion about what he screenPxRange is. You can rephrase some of my texts from the long comment

62_CAD/common.hlsl
```diff
@@ -197,6 +211,8 @@ NBL_CONSTEXPR uint32_t InvalidTextureIdx = nbl::hlsl::numeric_limits<uint32_t>::
...
+NBL_CONSTEXPR float MsdfSize = 32.0; // TODO: Should this be dynamic?
```
Erfan-Ahmadi, line 215: nope, but make sure to use MSDF all caps + use that on the c++ side as well

62_CAD/fragment_shader.hlsl
```diff
@@ -372,6 +372,23 @@ float32_t4 calculateFinalColor<true>(const uint2 fragCoord, const float localAlp
...
+    float2 unitRange = MsdfPixelRange / distanceFieldSizePixels;
+    return max(0.5 * dot(unitRange, screenSpaceSizePixels), 1.0);
+}
+
+float msdfOpacity(float3 msd, float2 screenPxRangeValue) {
+    float sd = median(msd.r, msd.g, msd.b);
+    float screenPxDistance = screenPxRangeValue * (sd - 0.5);
+    float opacity = clamp(screenPxDistance + 0.5, 0.0, 1.0);
```
Erfan-Ahmadi, line 388: Don't output opacity, output screen space distances and do AA with smooth step like we adjust do.
This is literally a sdf function, but reads from texture

deprilula28: Done

62_CAD/common.hlsl
deprilula28: Done

62_CAD/fragment_shader.hlsl
deprilula28: Done

62_CAD/vertex_shader.hlsl
deprilula28: Done

PR #117, opened 2024-06-03
Compute shader Merge Sort implementation

rtarun9:
## Description
This PR adds a compute shader based merge sort example. The shader uses shared memory (as much as possible) for higher performance.

## Testing
The program randomly fills a buffer (whose size is defined by the constant `NumberOfElementsToSort`.
The random numbers are in the range -INT_MIN to INT_MAX. Once the shader has executed, I sort teh input buffer on the CPU side (using std::sort) and use the following condition to determine if the output is correct or not:

```cpp
for (auto i = 0; i < NumberOfElementsToSort; i++)
{
    printf("output buffer -> %d input buffer -> %d\n", outputBufferData[i], inputBufferData[i]);
    if (outputBufferData[i] != inputBufferData[i])
    {
        return logFail("%d != %d\n", outputBufferData[i], inputBufferData[i]);
    }
}
```
		
A few sample outputs with varying input sizes:
image
image
image

## Explanation of the algorithm

I've used a log2(N) algorithm, where N is the number of elements to be sorted. In each phase, the output and input get *reversed* (as this is a in-place sorting algorithm).

Once all the phases are complete, the algorithm will decide which buffer has the output -> the 'input' buffer, or the 'output' buffer (note that the name 'input' buffer is used to indicate which buffer is initially fed the input data).

I've used shared memory as well. Basically, I put *as much data* as possible into shared memory. In case the input data is large and more than the physical devices shared memory limit, I fetch as little data as possible from global memory.


CMakeLists.txt
```diff
@@ -62,5 +62,6 @@ if(NBL_BUILD_EXAMPLES)
 	add_subdirectory(61_UI EXCLUDE_FROM_ALL)
 	add_subdirectory(62_CAD EXCLUDE_FROM_ALL)
 	add_subdirectory(62_SchusslerTest EXCLUDE_FROM_ALL)
+	add_subdirectory(64_MergeSort EXCLUDE_FROM_ALL)
```
devshgraphicsprogramming, line 65: you can take nr 11 if you want

64_MergeSort/main.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+	// Semaphore is used because before completion of merge sort phase X, phase X + 1 cannot start.
+	smart_refctd_ptr<ISemaphore> m_phaseSemaphore{};
+
+	smart_refctd_ptr<IGPUDescriptorSetLayout> m_descriptorSetLayout{};
+
+	// There are 2 descriptor sets for merge sort. One has input buffer / output buffer at binding 0, 1, and one has the inverse.
+	// This is because in each merge sort phase, the inputs and output buffers are *swapped*.
+	smart_refctd_ptr<nbl::video::IGPUDescriptorSet> m_descriptorSet1{};
+	smart_refctd_ptr<nbl::video::IGPUDescriptorSet> m_descriptorSet2{};
```
devshgraphicsprogramming, line 360: you don't need any descriptor sets, just use BDA for the inputs

64_MergeSort/main.cpp
```diff
@@ -0,0 +1,368 @@
...
+		for (auto i = 0; i < NumberOfElementsToSort; i++)
+		{
+			printf("output buffer -> %d input buffer -> %d\n", outputBufferData[i], inputBufferData[i]);
+			if (outputBufferData[i] != inputBufferData[i])
+			{
+				return logFail("%d != %d\n", outputBufferData[i], inputBufferData[i]);
+			}
+		}
+
+		printf("Merge sort succesfull!\n");
```
devshgraphicsprogramming, line 328: use the `m_logger` or whatever you base app has, then all messagess go to same output

64_MergeSort/main.cpp
```diff
@@ -0,0 +1,368 @@
...
+		}
+
+		// Now, get the current pointer to output buffer (the output may be in either input or output buffer).
+		// Perform merge sort on the CPU to test whether GPU compute version computed the result correctly.
+		auto outputBufferData = reinterpret_cast<const int32_t*>(m_outputBufferAllocation.memory->getMappedPointer());
+
+		if (numberOfPhases % 2 == 0)
+		{
+			outputBufferData = reinterpret_cast<const int32_t*>(m_inputBufferAllocation.memory->getMappedPointer());
+		}
```
devshgraphicsprogramming, line 314: well don't call them input and output then, just call them buffer A and B

64_MergeSort/main.cpp
```diff
@@ -0,0 +1,368 @@
...
+			};
+
+			const SortingPhaseData pushConstantData = {
+				.numElementsPerArray = (uint32_t)pow(2, phaseIndex - 1),
+				.bufferLength = static_cast<uint32_t>(NumberOfElementsToSort),
+			};
+
+			cmdbuf->pushConstants(m_pipelineLayout.get(), nbl::asset::IShader::ESS_COMPUTE, 0u, sizeof(pushConstantData), &pushConstantData);
+
+			cmdbuf->dispatch((uint32_t)ceilf(NumberOfElementsToSort / powf(2.0f, phaseIndex)), 1, 1);
```
devshgraphicsprogramming, line 276: why areyou using floats to do integer math !?

64_MergeSort/main.cpp
```diff
@@ -0,0 +1,368 @@
...
+				queue->submit(submitInfos);
+				queue->endCapture();
+
+			}
+			const ISemaphore::SWaitInfo wait_infos[] = { {
+		.semaphore = m_phaseSemaphore.get(),
+		.value = monotonicallyIncreasingCounter,
+	}, };
+
+			m_device->blockForSemaphores(wait_infos);
```
devshgraphicsprogramming, line 304: use frames in flight properly, there's no need to make the CPU block... just create log2(N) commandbuffers you don't recycle

and use BDA, not descriptor sets then you don't need log2(N) descriptor sets

devshgraphicsprogramming: actually, don't use log2(N) submits, use a single commandbuffer and dispatch

use pipeline barriers to synchronize between subsequent compute dispatches

64_MergeSort/main.cpp
```diff
@@ -0,0 +1,368 @@
...
+		static constexpr uint32_t WorkgroupSize = 64;
+		static constexpr uint32_t NumberOfElementsToSort = 1024 * 633u;
+		static const uint32_t WorkgroupCount = (uint32_t)ceil(NumberOfElementsToSort / (float)WorkgroupSize);
+
+		const size_t bufferSize = sizeof(int32_t) * NumberOfElementsToSort;
+
+		std::vector<int32_t> inputBufferData(NumberOfElementsToSort);
+		{
+			// Setup the input data (random) that is to be sorted.
+			const auto seed = std::chrono::system_clock::now().time_since_epoch().count();
```
devshgraphicsprogramming, line 50: use a fixed seed, it makes CI a nightmare to use a variable one

64_MergeSort/main.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+		// Get the maxinum bytes of shared memory we can use.
+		// In case the number of elements shared memory array can hold exceeds the limit, then switch to fetching data from global memory directly (for as few memory elements as possible).
+		const auto physicalDeviceLimits = m_device->getPhysicalDevice()->getLimits();
+		const uint32_t maxNumberOfArrayElementsSharedMemoryCanHold = physicalDeviceLimits.maxComputeSharedMemorySize / sizeof(int);
+
+		// Compute and input related constants.
+		static constexpr uint32_t WorkgroupSize = 64;
+		static constexpr uint32_t NumberOfElementsToSort = 1024 * 633u;
+		static const uint32_t WorkgroupCount = (uint32_t)ceil(NumberOfElementsToSort / (float)WorkgroupSize);
```
devshgraphicsprogramming, line 43: just stop it, stop using floats for integer math

64_MergeSort/main.cpp
```diff
@@ -0,0 +1,368 @@
...
+		}
+		if (!asset_base_t::onAppInitialized(std::move(system)))
+		{
+			return false;
+		}
+
+		// Get the maxinum bytes of shared memory we can use.
+		// In case the number of elements shared memory array can hold exceeds the limit, then switch to fetching data from global memory directly (for as few memory elements as possible).
+		const auto physicalDeviceLimits = m_device->getPhysicalDevice()->getLimits();
+		const uint32_t maxNumberOfArrayElementsSharedMemoryCanHold = physicalDeviceLimits.maxComputeSharedMemorySize / sizeof(int);
```
devshgraphicsprogramming, line 38: > I've used shared memory as well. Basically, I put as much data as possible into shared memory. In case the input data is large and more than the physical devices shared memory limit, I fetch as little data as possible from global memory.

You get optimum residency while using about 16 to 32 bytes of smem per workgroup invocation, just make that a tweakable parameter.

If you end up using MAX then you end up with a occupancy problem (see CUDA occupancy calculator)

64_MergeSort/main.cpp
```diff
@@ -0,0 +1,368 @@
...
+using namespace nbl;
+using namespace core;
+using namespace system;
+using namespace asset;
+using namespace video;
+
+// A high level overview of the algorithm implementation (merge sort).
+// This will be a X pass algorithm, where X is log2(number_of_elements).
+// In each phase, you have to 'merge' 2 sorted list of elements into a single list.
+// Each of these 'lists' will have thier number of elements roughly doubled in each phase (after the merge step of previous phase).
```
devshgraphicsprogramming, line 13: I was kinda expecting you to roll up the CPU logic into a separate class + shared HLSL/c++ header to drive it (see the counting sort code)

64_MergeSort/app_resources/merge_sort.hlsl
```diff
@@ -0,0 +1,77 @@
...
+[[vk::binding(1,0)]] RWStructuredBuffer<int> input_buffer;
+
+struct SortingPhaseData
+{
+    uint num_elements_per_array;
+    uint buffer_length;
+};
+
+// Rather than indexing into the input array directly, use group shared memory.
+groupshared int shared_memory_input_array[MaxNumberOfArrayElementsSharedMemoryCanHold];
```
devshgraphicsprogramming, line 13: I'd make a separate HLSL header like counting sort to shove all the logic into, also use the Accessor Pattern to make the algorithm independent of the resource thats providing the inputs/outputs to the algo

64_MergeSort/main.cpp
devshgraphicsprogramming: Also use virtual (persistent) workgroups, so you `max()` the workgroup count against a device specific limit (`SPhysicalDeviceLimits` has a helper func for this)

64_MergeSort/app_resources/merge_sort.hlsl
devshgraphicsprogramming: Also you know, using a template let us sort something other than `int` (as long as it has an `operator<`

64_MergeSort/app_resources/merge_sort.hlsl
```diff
@@ -0,0 +1,77 @@
...
+    {
+        right_array_end = phase_data.buffer_length - 1;
+    }
+
+    // Now that the left and right array bounds are determined, move the data into shared memory.
+    // If there is some *excess* data from the input buffer we cannot move to shared memory, store as much as data in shared memory as possible and fetch the rest from global memory.
+    for (uint i = left_array_start; i <= min(right_array_end, MaxNumberOfArrayElementsSharedMemoryCanHold - 1); i++)
+    {
+        shared_memory_input_array[i] = input_buffer[i];
+    }
```
devshgraphicsprogramming, line 40: load the elements cooperatively so you don't get bank-conflicts

64_MergeSort/app_resources/merge_sort.hlsl
```diff
@@ -0,0 +1,77 @@
...
+        output_buffer[index++] = left_array_current_element;
+        ++left_array_start;
+    }
+
+    while (right_array_start <= right_array_end)
+    {
+        int right_array_current_element = (right_array_start < MaxNumberOfArrayElementsSharedMemoryCanHold) ? shared_memory_input_array[right_array_start] : input_buffer[right_array_start];
+        output_buffer[index++] = right_array_current_element;
+        ++right_array_start;
+    }
```
devshgraphicsprogramming, line 76: this drops to 0 parallelism as your segment lengths increase

Almost totally wrong, please read: https://moderngpu.github.io/mergesort.html

CMakeLists.txt
rtarun9: This has been done, merge sort is now in example 11 over example 64.

64_MergeSort/main.cpp
rtarun9: Done, for getting integer ceil division of a / b I'm now using (a + b - 1) / b instead.

64_MergeSort/main.cpp
rtarun9: Done, seed is not hardcoded.

64_MergeSort/main.cpp
rtarun9: Got it, will use BDA over descriptor sets.

64_MergeSort/main.cpp
rtarun9: Done, will stop using printf in the future and use m_logger.

64_MergeSort/main.cpp
rtarun9: Renamed input / output buffer to bufferA / bufferB.

PR #150, opened 2024-10-07
Ef64 benchmark

Przemog1:


64_EmulatedFloatTest/app_resources/benchmark/benchmark.comp.hlsl
```diff
@@ -0,0 +1,144 @@
...
+
+#include "app_resources/benchmark/common.hlsl"
+#include <nbl/builtin/hlsl/emulated/float64_t.hlsl>
+#include <nbl/builtin/hlsl/random/xoroshiro.hlsl>
+#include <nbl/builtin/hlsl/math/quadrature/gauss_legendre/gauss_legendre.hlsl>
+#include <nbl/builtin/hlsl/ieee754.hlsl>
+
+using namespace nbl::hlsl;
+
+RWByteAddressBuffer outputBuffer;
```
Erfan-Ahmadi, line 14: explicitly write `[[vk::binding(0, 0)]] `

64_EmulatedFloatTest/main.cpp
```diff
@@ -1033,6 +1047,292 @@ class CompatibilityTest final : public MonoDeviceApplication, public MonoAssetMa
...
+
+                    dummyBuff->setObjectDebugName("benchmark buffer");
+
+                    nbl::video::IDeviceMemoryBacked::SDeviceMemoryRequirements reqs = dummyBuff->getMemoryReqs();
+
+                    m_allocation = base.m_device->allocate(reqs, dummyBuff.get(), nbl::video::IDeviceMemoryAllocation::EMAF_NONE);
+                    if (!m_allocation.isValid())
+                        base.logFail("Failed to allocate Device Memory compatible with our GPU Buffer!\n");
+
+                    assert(dummyBuff->getDeviceAddress() < std::numeric_limits<uint32_t>::max()); // TODO: is it even possible?
```
Erfan-Ahmadi, line 1165: buffer device address is 64bits, why are you trying to clamp it to 32? It's risky.

64_EmulatedFloatTest/main.cpp
```diff
@@ -1033,6 +1047,292 @@ class CompatibilityTest final : public MonoDeviceApplication, public MonoAssetMa
...
+            m_device->waitIdle();
+
+            beforeTimestampSignals[0].value = ++m_beforeTimestampSemaphoreCounter;
+            afterTimestapSignals[0].value = ++m_afterTimestampSemaphoreCounter;
+
+            const uint64_t nativeBenchmarkTimeElapsedNanoseconds = calcTimeElapsed();
+            const float nativeBenchmarkTimeElapsedSeconds = double(nativeBenchmarkTimeElapsedNanoseconds) / 1000000000.0;
+
+            m_logger->log("%llu ns, %f s", ILogger::ELL_PERFORMANCE, nativeBenchmarkTimeElapsedNanoseconds, nativeBenchmarkTimeElapsedSeconds);
+        }
```
Erfan-Ahmadi, line 1267: If I understood  correctly you're having 3 command buffers (beforeTimeStamp, MainWork, afterTimestamp) and you're submitting them and ensuring the order by waitIdle instead of semaphores.
and You're submitting the MainWork Iteration Times (1000), It makes sense to do that instead of recording a command buffer with 1000 dispatches inside, I get that you also want to keep reusing your command buffer.

waitIdle syncs with CPU and then submits back to GPU, to get better timings you need to extract your first and last  mainWork submits and connect then to the timeStamp submits via semaphores instead of waitIdle, so first submit wait for beforeTimestamp, and afterTimestamp waits for last submit. this should be easy to achieve by doing you loop Iterations-1 times, and writing 2 lines of waitSemaphore codes 🙂

64_EmulatedFloatTest/main.cpp
Erfan-Ahmadi: oh shit, you're using byteAddressBuffer not BDA, you shouldn't be using getDeviceAddress for that, your offset is just 0 in this case 😄
I know it was for test but you could definitely crash your gpu doing that. cause it wayyy out of bound 😄

64_EmulatedFloatTest/app_resources/benchmark/common.hlsl
```diff
@@ -0,0 +1,24 @@
+//// Copyright (C) 2023-2024 - DevSH Graphics Programming Sp. z O.O.
+//// This file is part of the "Nabla Engine".
+//// For conditions of distribution and use, see copyright notice in nabla.h
+
+#include <nbl/builtin/hlsl/cpp_compat.hlsl>
+
+NBL_CONSTEXPR uint32_t BENCHMARK_WORKGROUP_DIMENSION_SIZE_X = 128u;
+NBL_CONSTEXPR uint32_t BENCHMARK_WORKGROUP_DIMENSION_SIZE_Y = 1u;
+NBL_CONSTEXPR uint32_t BENCHMARK_WORKGROUP_DIMENSION_SIZE_Z = 1u;
+NBL_CONSTEXPR uint32_t BENCHMARK_WORKGROUP_SIZE = 1024u;
```
Erfan-Ahmadi, line 10: BENCHMARK_WORKGROUP_COUNT

PR #156, opened 2024-10-30
Cameraz

devshgraphicsprogramming:
Makes 61_UI consume Nabla::ext::Cameras as a downstream example and harness.


common/include/ICamera.hpp
```diff
@@ -0,0 +1,82 @@
...
+	struct Traits
+	{
+		using controller_t = base_t;
+		using projection_t = typename controller_t::projection_t;
+		using gimbal_t = typename controller_t::CGimbal;
+		using controller_virtual_event_t = typename controller_t::CVirtualEvent;
+	};
+
+	ICamera(core::smart_refctd_ptr<typename Traits::gimbal_t>&& gimbal, core::smart_refctd_ptr<typename Traits::projection_t> projection, const float32_t3& target = {0,0,0})
+		: base_t(core::smart_refctd_ptr(gimbal)), m_projection(core::smart_refctd_ptr(projection)) { setTarget(target); }
```
devshgraphicsprogramming, line 34: not sure that we want gimbals to be freely allocatable on the heap and refcounted, I'm thinking that camera should OWN them and they should live inside the camera as a member (and not inherit from IReferenceCounted)

common/include/ICamera.hpp
```diff
@@ -0,0 +1,82 @@
...
+	{
+		m_target = position;
+	
+		const auto* gimbal = base_t::m_gimbal.get();
+		auto localTarget = m_target - gimbal->getPosition();
+
+		// TODO: use gimbal to perform a rotation!
+
+		recomputeViewMatrix();
+	}
```
devshgraphicsprogramming, line 54: may I suggest a trick to avoid recomputing view matrix all the time?

keep a "dirty flag" / incrementing counter in the gimbals that increments after every gimbal manipulation.

then make the camera track the last counter values of all gimbals during `getViewMatrix`, then you know if the view matrix is stale or not and whether to call `recomputeViewMatrix`

devshgraphicsprogramming: Also the position and target setting should happen purely based on gimbal manipulation, its a bit of code duplication and inversion of control for anyone to do it through the camera

common/include/CCamera.hpp
```diff
@@ -2,325 +2,94 @@
...
-	bool leftHanded, firstUpdate = true, mouseDown = false;
-	
-	std::array<nbl::ui::E_KEY_CODE, ECMK_COUNT> keysMap = { {nbl::ui::EKC_NONE} }; // map camera E_CAMERA_MOVE_KEYS to corresponding Nabla key codes, by default camera uses WSAD to move
-	// TODO: make them use std::array
-	bool keysDown[E_CAMERA_MOVE_KEYS::ECMK_COUNT] = {};
-	double perActionDt[E_CAMERA_MOVE_KEYS::ECMK_COUNT] = {}; // durations for which the key was being held down from lastVirtualUpTimeStamp(=last "guessed" presentation time) to nextPresentationTimeStamp
-
-	std::chrono::microseconds nextPresentationTimeStamp, lastVirtualUpTimeStamp;
+    virtual void manipulate(std::span<const typename traits_t::controller_virtual_event_t> virtualEvents) override
+    {
```
devshgraphicsprogramming, line 30: I guess CCamera needs to be called a CFPSCamera then?

common/include/camera/CCubeProjection.hpp
```diff
@@ -0,0 +1,36 @@
...
+struct CCubeProjectionConstraints
+{
+    using matrix_t = typename T;
+    using projection_t = typename IProjection<typename matrix_t>;
+    using projection_range_t = std::array<typename core::smart_refctd_ptr<projection_t>, 6u>;
+    using base_t = ILinearProjection<typename projection_range_t>;
+};
+
+//! Class providing linear cube projections with projection matrix per face of a cube, each projection matrix represents a single view-port
+template<ProjectionMatrix T = float64_t4x4>
```
devshgraphicsprogramming, line 19: you could actually always make the projection and view matrices 64bit, there's no downsides to that

common/include/ICamera.hpp
```diff
@@ -0,0 +1,82 @@
...
+		const auto* gimbal = base_t::m_gimbal.get();
+		const auto& position = gimbal->getPosition();
+
+		const auto [xaxis, yaxis, zaxis] = std::make_tuple(gimbal->getXAxis(), gimbal->getYAxis(), gimbal->getZAxis());
+		m_viewMatrix[0u] = float32_t4(xaxis, -hlsl::dot(xaxis, position));
+		m_viewMatrix[1u] = float32_t4(yaxis, -hlsl::dot(yaxis, position));
+		m_viewMatrix[2u] = float32_t4(zaxis, -hlsl::dot(zaxis, position));
+	}
+
+	const core::smart_refctd_ptr<typename Traits::projection_t> m_projection;
```
devshgraphicsprogramming, line 75: have a think whether its better to have the projection own the camera and not the camera own the projection, why?

projection needs to produce matrices/functions based off the view matrix, and you know there will only ever be one.

If you do the reverse then the projection can't know about the camera, which is rather difficult for e.g. a cubemap to produce 6 concatenated MVP matrices for each face

common/include/camera/ICameraControl.hpp
```diff
@@ -0,0 +1,327 @@
...
+
+            assert(isOrthoBase(xaxis, yaxis, zaxis));
+        }
+
+        float32_t3 m_position;
+        glm::quat m_orientation;
+
+        // Represents the camera's orthonormal basis
+        // https://en.wikipedia.org/wiki/Orthonormal_basis
+        float32_t3x3 m_orthonormal;
```
devshgraphicsprogramming, line 158: only keep the quaternion, and spit out the matrix on demand, its not that expensive to compute it

common/include/camera/ICameraControl.hpp
```diff
@@ -0,0 +1,327 @@
...
+
+        inline void move(float32_t3 delta)
+        {
+            m_position += delta;
+        }
+
+        inline void reset()
+        {
+            // TODO
+        }
```
devshgraphicsprogramming, line 106: not sure you want this (in the gimbal that is, the camera can have it, e.g. reference/strating gimbal positions)

common/include/camera/ICameraControl.hpp
```diff
@@ -0,0 +1,327 @@
...
+
+        inline void advance(float distance)
+        {
+            move({ distance, 0.f, 0.f });
+        }
+
+        inline void move(float32_t3 delta)
+        {
+            m_position += delta;
+        }
```
devshgraphicsprogramming: I'm not sure these make sense in a global coordinate system, they'd always have to be relative to the local coordiate frame described by the quaternion.

This leads me to a question, is your `m_position` global or relative (happens before or after the rotation)

common/include/camera/ICameraControl.hpp
```diff
@@ -0,0 +1,312 @@
...
+
+        // Position of gimbal
+        inline const float32_t3& getPosition() const { return m_position; }
+
+        // Orientation of gimbal
+        inline const glm::quat& getOrientation() const { return m_orientation; }
+
+        // Orthonormal [getXAxis(), getYAxis(), getZAxis()] matrix
+        inline const float32_t3x3& getOrthonornalMatrix() const { return m_orthonormal; }
```
devshgraphicsprogramming, line 115: to avoid semantical confusion (as to what your postion means) I'd also add a `getMatrix()` method which would return a `float64_t3x4` (a full TRS).

P.s. if you like you can even make it an `operator float64_t3x4`

common/include/camera/ICameraControl.hpp
devshgraphicsprogramming: Also keep everything as `float64_t`

common/include/camera/ICameraControl.hpp
```diff
@@ -0,0 +1,312 @@
...
+
+            using virtual_key_state_t = std::tuple<ui::E_KEY_CODE /*physical key representing virtual key*/, bool /*is pressed*/, float64_t /*delta action*/>;
+
+            auto updateVirtualState = [&]() -> virtual_key_state_t
+            {
+                virtual_key_state_t state = { ui::E_KEY_CODE::EKC_NONE, false, 0.f };
+
+                for (const auto& ev : events) // TODO: improve the search
+                {
+                    if (ev.keyCode == code)
```
devshgraphicsprogramming, line 213: maintain a simple table of all key codes to actions/virtual keys (and maybe even callbacks/lambdas to execute on them)

common/include/camera/ICameraControl.hpp
```diff
@@ -0,0 +1,312 @@
...
+        EventsCount
+    };
+
+    //! Virtual event representing a manipulation
+    struct CVirtualEvent
+    {
+        using manipulation_encode_t = float64_t;
+
+        VirtualEventType type;
+        manipulation_encode_t value;
```
devshgraphicsprogramming, line 71: can call it `magnitude` for clarity if you want

common/include/camera/ICameraControl.hpp
```diff
@@ -0,0 +1,312 @@
...
+    };
+
+    class CGimbal : virtual public core::IReferenceCounted
+    {
+    public:
```
devshgraphicsprogramming, line 76: not sure a gimbal needs to be refcounted OR be nested in the ICameraController class

common/include/camera/ICameraControl.hpp
devshgraphicsprogramming: also wouldn't the events belong to an `IGimbal` because they're for gimbals?

common/include/camera/ICameraControl.hpp
```diff
@@ -0,0 +1,312 @@
...
+    {
+        m_nextPresentationTimeStamp = nextPresentationTimeStamp;
+        return;
+    }
+
+    // manipulate camera with gimbal & virtual events, begin must be called before that!
+    virtual void manipulate(std::span<const CVirtualEvent> virtualEvents) = 0;
+
+    // finish controller manipulation session, call after last manipulate in the hot loop
+    void end(std::chrono::microseconds nextPresentationTimeStamp)
```
devshgraphicsprogramming, line 180: the begin/end don't actually affect `manipulate` because the events already have a magnitude/impulse by them.

its actually event  generation/processing/translation that youre beginning/ending that need and updates the timestamps

common/include/camera/ICameraControl.hpp
```diff
@@ -0,0 +1,312 @@
...
+        }
+
+        return output;
+    }
+
+    inline void setMoveSpeed(const float moveSpeed) { m_moveSpeed = moveSpeed; }
+    inline void setRotateSpeed(const float rotateSpeed) { m_rotateSpeed = rotateSpeed; }
+
+    inline const float getMoveSpeed() const { return m_moveSpeed; }
+    inline const float getRotateSpeed() const { return m_rotateSpeed; }
```
devshgraphicsprogramming, line 281: not sure they belong in the "camera interface for all cameras"

common/include/camera/ICameraControl.hpp
```diff
@@ -0,0 +1,312 @@
...
+            pitch.type = dPitch > 0.f ? TiltUp : TiltDown;
+            pitch.value = std::abs(dPitch);
+        }
+
+        if (dYaw)
+        {
+            auto& yaw = output.emplace_back();
+            yaw.type = dYaw > 0.f ? PanRight : PanLeft;
+            yaw.value = std::abs(dYaw);
+        }
```
devshgraphicsprogramming, line 272: this seems very "hardwired" to how FPS camera works

common/include/camera/ICameraControl.hpp
```diff
@@ -0,0 +1,312 @@
...
+        m_keysToEvent[TiltUp] =  ui::E_KEY_CODE::EKC_NONE ;
+        m_keysToEvent[TiltDown] =  ui::E_KEY_CODE::EKC_NONE ;
+        m_keysToEvent[PanLeft] =  ui::E_KEY_CODE::EKC_NONE ;
+        m_keysToEvent[PanRight] =  ui::E_KEY_CODE::EKC_NONE ;
+        m_keysToEvent[RollLeft] =  ui::E_KEY_CODE::EKC_NONE ;
+        m_keysToEvent[RollRight] =  ui::E_KEY_CODE::EKC_NONE ;
+        m_keysToEvent[Reset] =  ui::E_KEY_CODE::EKC_R ;
+    }
+
+    core::smart_refctd_ptr<CGimbal> m_gimbal;
```
devshgraphicsprogramming, line 302: don't make the gimbal live in the main camera interface, some cameras may have multiple gimbals

common/include/camera/IProjection.hpp
```diff
@@ -0,0 +1,67 @@
...
+    inline const value_t& getMatrix() { return m_projectionMatrix; }
+    inline bool isLeftHanded() { return m_isLeftHanded; }
+
+private:
+    inline void updateHandnessState()
+    {
+        m_isLeftHanded = hlsl::determinant(m_projectionMatrix) < 0.f;
+    }
+
+    value_t m_projectionMatrix;
```
devshgraphicsprogramming, line 36: this belongs in `LinearProjection`, the basic `IProjection` should be for anything, thin-lens, spherical, octahedral, etc.

common/include/camera/ILinearProjection.hpp
```diff
@@ -0,0 +1,29 @@
...
+    using range_t = typename base_t::range_t;
+    using projection_t = typename base_t::projection_t;
+
+    ILinearProjection(range_t&& projections) : base_t(std::move(projections)) {}
+
+protected:
+    inline range_t& getViewportProjections()
+    {
+        return base_t::m_projectionRange;
+    }
```
devshgraphicsprogramming, line 24: it would be good to have this own/reference the camera so it can pull the view matrix and be ready to spit out MVP, MV, P, P-1, MV-1, MVP-1 matrices for you

common/include/camera/ICameraControl.hpp
AnastaZIuk: removed

common/include/camera/ICameraControl.hpp
AnastaZIuk: resolved

common/include/camera/ICameraControl.hpp
AnastaZIuk: > can call it magnitude for clarity if you want

resolved

> also wouldn't the events belong to an IGimbal because they're for gimbals?

if we have generic virtual camera events like now then I would say no, also what `IGimbal` would be? Do we expect any derived classes for which it would be base? Imo any type of camera can be currently implemented with `CGimbal` instances, so for me `IGimbal` doesn't make much sense at least I don't see what we would store there. I have also moved events to new `CVirtualGimbalEvent` outside templates in recent commit

common/include/camera/ICameraControl.hpp
AnastaZIuk: none of that, derived class implements `getGimbal` which models camera view (and there is always only one such gimbal regardless the type of camera we talk about), we don't make any assumptions now & can easily store any range of gimbals if we expect multiple under the hood (follow camera)

common/include/camera/ICameraControl.hpp
AnastaZIuk: resolved

common/include/camera/ICameraControl.hpp
AnastaZIuk: removed

common/include/camera/ICameraControl.hpp
AnastaZIuk: the semantic indeed suggest FPS camera like stuff, however the impulses we return can be used to implement any type of camera, need to think about this but mouse events at the end are just XY delta motion

common/include/camera/ICameraControl.hpp
AnastaZIuk: resolved

common/include/camera/ICameraControl.hpp
AnastaZIuk: actually let's create IGimbal which is interface for any object not just camera and add scale events

devshgraphicsprogramming:
btw why are you adding to ex 61, it would be nice if cameras had their own thing

common/include/camera/ICameraControl.hpp
AnastaZIuk: resolved

PR #157, opened 2024-11-05
Ray query geometry creator

devshgraphicsprogramming:


67_RayQueryGeometry/common.hlsl
```diff
@@ -0,0 +1,25 @@
...
+NBL_CONSTEXPR uint32_t WorkgroupSize = 16;
+
+#ifdef __HLSL_VERSION
+struct VertexData
+{
+    float3 position;
+    float4 color;
+    float2 uv;
+    float3 normal;
+};
```
devshgraphicsprogramming, line 13: with modern rendering you REALLY don't want to use interleaved vertex formats

67_RayQueryGeometry/common.hlsl
```diff
@@ -0,0 +1,25 @@
...
+
+struct SCameraParameters //! matches CPU version size & alignment (160, 4)
+{
+    float3 camPos;
+
+	float4x4 MVP;
+	float3x4 V;
+    float3x4 P;
+};
```
devshgraphicsprogramming, line 22: this struct can be shared between HLSL and C++ as long as you `using namespace nbl::hlsl` beforehand

67_RayQueryGeometry/main.cpp
```diff
@@ -0,0 +1,1010 @@
...
+					.mayAlias = false
+				},
+			/*.loadOp = */{IGPURenderpass::LOAD_OP::CLEAR},
+			/*.storeOp = */{IGPURenderpass::STORE_OP::STORE},
+			/*.initialLayout = */{IGPUImage::LAYOUT::UNDEFINED}, // because we clear we don't care about contents
+			/*.finalLayout = */{IGPUImage::LAYOUT::ATTACHMENT_OPTIMAL} // transition to presentation right away so we can skip a barrier
+		}},
+		IGPURenderpass::SCreationParams::DepthStencilAttachmentsEnd
+		};
+		m_params.depthStencilAttachments = depthAttachments;
```
devshgraphicsprogramming, line 44: you don't need a depth buffer for this example, see example 27

67_RayQueryGeometry/main.cpp
```diff
@@ -0,0 +1,1010 @@
...
+#include "common.hpp"
+
+struct SCameraParameters
+{
+	float camPos[3];
+
+	float MVP[4 * 4];
+	float V[3 * 4];
+	float P[3 * 4];
+};
```
devshgraphicsprogramming, line 14: https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/157/files#r1829166059

67_RayQueryGeometry/main.cpp
```diff
@@ -0,0 +1,1010 @@
...
+					// last place where the depth can get modified
+					.srcStageMask = PIPELINE_STAGE_FLAGS::COLOR_ATTACHMENT_OUTPUT_BIT,
+					// only write ops, reads can't be made available
+					.srcAccessMask = ACCESS_FLAGS::COLOR_ATTACHMENT_WRITE_BIT
+					// spec says nothing is needed when presentation is the destination
+				}
+				// leave view offsets and flags default
+			},
+			IGPURenderpass::SCreationParams::DependenciesEnd
+			};
```
devshgraphicsprogramming, line 195: you won't even have a rnederpass in this example

67_RayQueryGeometry/main.cpp
```diff
@@ -0,0 +1,1010 @@
...
+						.stageFlags = asset::IShader::E_SHADER_STAGE::ESS_COMPUTE,
+						.count = 1,
+					},
+					{
+						.binding = 1,
+						.type = asset::IDescriptor::E_TYPE::ET_UNIFORM_BUFFER,
+						.createFlags = IGPUDescriptorSetLayout::SBinding::E_CREATE_FLAGS::ECF_NONE,
+						.stageFlags = asset::IShader::E_SHADER_STAGE::ESS_COMPUTE,
+						.count = 1,
+					},
```
devshgraphicsprogramming, line 362: if your uniforms can fit in push constants, don't complicate your life with UBOs

67_RayQueryGeometry/main.cpp
```diff
@@ -0,0 +1,1010 @@
...
+		}
+
+		inline void workLoopBody() override
+		{
+			const auto resourceIx = m_realFrameIx % m_maxFramesInFlight;
+
+			if (m_realFrameIx >= m_maxFramesInFlight)
+			{
+				const ISemaphore::SWaitInfo cbDonePending[] =
+				{
```
devshgraphicsprogramming, line 425: watch for what @Erfan-Ahmadi is doing on `master` branch and apply a similar change here after merging with `master`

67_RayQueryGeometry/main.cpp
```diff
@@ -0,0 +1,1010 @@
...
+			memcpy(uboData.P, projectionMatrix.pointer(), sizeof(uboData.P));
+			{
+				SBufferRange<IGPUBuffer> range;
+				range.buffer = smart_refctd_ptr(cameraBuffer);
+				range.size = cameraBuffer->getSize();
+
+				cmdbuf->updateBuffer(range, &uboData);
+			}
+
+			auto* queue = getGraphicsQueue();
```
devshgraphicsprogramming, line 499: @AnastaZIuk after your cameras get merged, advise on changes here

67_RayQueryGeometry/main.cpp
```diff
@@ -0,0 +1,1010 @@
...
+						else
+							--m_realFrameIx;
+					}
+				}
+
+				std::string caption = "[Nabla Engine] Geometry Creator";
+				{
+					caption += ", displaying [" + std::string(object.meta.name.data()) + "]";
+					m_window->setCaption(caption);
+				}
```
devshgraphicsprogramming, line 682: take the choice of what to display away, display all geometries in a line side-by-side (one TLAS, multiple BLASes - one per geo)

67_RayQueryGeometry/main.cpp
```diff
@@ -0,0 +1,1010 @@
...
+			copyInfo.dst = gpuBlas.get();
+			copyInfo.mode = IGPUBottomLevelAccelerationStructure::COPY_MODE::COMPACT;
+			cmdbuf->copyAccelerationStructure(copyInfo);
+
+			// build top level AS
+			{
+				const uint32_t instancesCount = 1;	// TODO: temporary for now
+				IGPUTopLevelAccelerationStructure::DeviceInstance instances[instancesCount];
+				core::matrix3x4SIMD identity;
+				instances[0].blas.deviceAddress = gpuBlas->getCreationParams().bufferRange.buffer->getDeviceAddress();
```
devshgraphicsprogramming, line 904: the TLAS references a BLAS and cannot happen at the same time as the BLAS build (or copies) of the BLASes its referencing, hence you need another pipeline barrier here.

67_RayQueryGeometry/main.cpp
```diff
@@ -0,0 +1,1010 @@
...
+			}
+
+			return true;
+		}
+
+		bool createAccelerationStructures(IGPUCommandBuffer* cmdbuf)
+		{
+			// build bottom level ASes
+			{
+				const auto& obj = objectsGpu[OT_CUBE];
```
devshgraphicsprogramming, line 792: when you progress to do all objects, do 3 waves (so you have most overlap/least pipeline barriers):
1. build all BLASes
2. compact all BLASes
3. build the TLAS

67_RayQueryGeometry/render.comp.hlsl
```diff
@@ -0,0 +1,77 @@
...
+
+    const float4x4 invMVP = inverse(cameraData.params.MVP);
+    
+    float4 NDC = float4(texCoords * float2(2.0, -2.0) + float2(-1.0, 1.0), 0.0, 1.0);
+    float3 targetPos;
+    {
+        vec4 tmp = invMVP * NDC;
+        targetPos = tmp.xyz / tmp.w;
+        NDC.z = 1.0;
+    }
```
devshgraphicsprogramming, line 43: compute all this on the CPU and send it as push constant to GPU

devshgraphicsprogramming: ofc I know `texCoords` is a variable but the whole `NDC` can be precomputed as `scale*threadID.xy+offset`

67_RayQueryGeometry/common.hlsl
devshgraphicsprogramming: hmmm I see that Geometry Creation spits out interleaved, in that case don't bother de-interleaving

67_RayQueryGeometry/app_resources/render.comp.hlsl
```diff
@@ -2,6 +2,7 @@
 
 struct SPushConstants
 {
+    bool useIndex;
     uint64_t vertexBufferAddress;
     uint64_t indexBufferAddress;
```
devshgraphicsprogramming, line 7: don't use `bool`s passing data around, you can e.g. check that `indexBufferAddress!=vertexBufferAddress` to see if you have an index buffer instead

67_RayQueryGeometry/main.cpp
```diff
@@ -887,36 +910,37 @@ class RayQueryGeometryApp final : public examples::SimpleWindowedApplication, pu
...
+			//IQueryPool::SCreationParams qParams{ .queryCount = 1, .queryType = IQueryPool::ACCELERATION_STRUCTURE_COMPACTED_SIZE };
+			//smart_refctd_ptr<IQueryPool> queryPool = m_device->createQueryPool(std::move(qParams));
+			//cmdbuf->resetQueryPool(queryPool.get(), 0, 1);
+
+			//uint32_t queryCount = 0;
+			//const IGPUAccelerationStructure* ases[1u] = { gpuBlas.get() };
+			//cmdbuf->writeAccelerationStructureProperties({ ases, 1}, IQueryPool::ACCELERATION_STRUCTURE_COMPACTED_SIZE, queryPool.get(), queryCount++);
+
+			//size_t asSizes[1];
+			//m_device->getQueryPoolResults(queryPool.get(), 0, queryCount, asSizes, sizeof(size_t), IQueryPool::WAIT_BIT);
```
devshgraphicsprogramming, line 922: you were using the query wrong, if something is a command buffer operation (`IGPUCommandBuffer` method) that means that it completes when the command buffer it was recorded into finishes execution.

You can only get the Query Pool Results AFTER the semaphore completes.

This would mean that you would need 2 separate command buffers and queue submits with the Host blocking after each:
1. build blas
2. compact blas & build tlas

the good news is that you don't need a pipeline barrier between 1 and 2 because they're separated by a timeline semaphore signal-wait pair (indirectly, via the host)

devshgraphicsprogramming: what you're doing now is actually causing a deadlock cause you're making the Host wait on the result of a query thats not even pending yet

devshgraphicsprogramming: also its not necessary to reset a query pool before using it (more like After0

67_RayQueryGeometry/main.cpp
```diff
@@ -0,0 +1,1088 @@
...
+					.baseMipLevel = 0u,
+					.levelCount = 1u,
+					.baseArrayLayer = 0u,
+					.layerCount = 1u
+				};
+				imageBarriers[0].oldLayout = IImage::LAYOUT::TRANSFER_DST_OPTIMAL;
+				imageBarriers[0].newLayout = IImage::LAYOUT::PRESENT_SRC;
+
+				cmdbuf->pipelineBarrier(E_DEPENDENCY_FLAGS::EDF_NONE, { .imgBarriers = imageBarriers });
+			}
```
devshgraphicsprogramming, line 614: btw example 27 stores straight to swapchain IIRC

67_RayQueryGeometry/main.cpp
```diff
@@ -0,0 +1,1088 @@
...
+				obj.bindings.vertex = { .offset = 0u, .buffer = std::move(vertexBuffer) };
+				SBufferRange<IGPUBuffer> vRange = { .offset = obj.bindings.vertex.offset, .size = obj.bindings.vertex.buffer->getSize(), .buffer = obj.bindings.vertex.buffer };
+				cmdbuf->updateBuffer(vRange, vBuffer->getPointer());
+
+				if (iBuffer)
+				{
+					obj.bindings.index = { .offset = 0u, .buffer = std::move(indexBuffer) };
+					SBufferRange<IGPUBuffer> iRange = { .offset = obj.bindings.index.offset, .size = obj.bindings.index.buffer->getSize(), .buffer = obj.bindings.index.buffer };
+					cmdbuf->updateBuffer(iRange, iBuffer->getPointer());
+				}
```
devshgraphicsprogramming, line 829: geometry creator may spit out a mesh thats too big to update a buffer for like that (e.g. more than 1200 vertices), use `IUtilities` to create or update the buffers with data instead (or asset converter!)

There's a 64kb limit to `cmdbuf->updateBuffer`

67_RayQueryGeometry/main.cpp
```diff
@@ -0,0 +1,1088 @@
...
+			{
+				SMemoryBarrier memBarrier;
+				memBarrier.srcStageMask = PIPELINE_STAGE_FLAGS::ACCELERATION_STRUCTURE_BUILD_BIT;
+				memBarrier.srcAccessMask = ACCESS_FLAGS::ACCELERATION_STRUCTURE_WRITE_BIT;
+				memBarrier.dstStageMask = PIPELINE_STAGE_FLAGS::ACCELERATION_STRUCTURE_BUILD_BIT;
+				memBarrier.dstAccessMask = ACCESS_FLAGS::ACCELERATION_STRUCTURE_READ_BIT;
+				cmdbufBlas->pipelineBarrier(E_DEPENDENCY_FLAGS::EDF_NONE, { .memBarriers = {&memBarrier, 1} });
+			}
+
+			cmdbufBlas->resetQueryPool(queryPool.get(), 0, 1);
```
devshgraphicsprogramming, line 926: why are you still resetting the query pool ?

67_RayQueryGeometry/main.cpp
```diff
@@ -0,0 +1,1088 @@
...
+				pRangeInfos[0] = &buildRangeInfos[0];
+
+				cmdbufBlas->buildAccelerationStructures({ &blasBuildInfo, 1 }, pRangeInfos);
+			}
+
+			{
+				SMemoryBarrier memBarrier;
+				memBarrier.srcStageMask = PIPELINE_STAGE_FLAGS::ACCELERATION_STRUCTURE_BUILD_BIT;
+				memBarrier.srcAccessMask = ACCESS_FLAGS::ACCELERATION_STRUCTURE_WRITE_BIT;
+				memBarrier.dstStageMask = PIPELINE_STAGE_FLAGS::ACCELERATION_STRUCTURE_BUILD_BIT;
```
devshgraphicsprogramming, line 921: merge master, the query stage is COPY not build

67_RayQueryGeometry/main.cpp
```diff
@@ -0,0 +1,1088 @@
...
+
+			cmdbufBlas->resetQueryPool(queryPool.get(), 0, 1);
+
+			uint32_t queryCount = 0;
+			const IGPUAccelerationStructure* ases[1u] = { gpuBlas.get() };
+			cmdbufBlas->writeAccelerationStructureProperties({ ases, 1}, IQueryPool::ACCELERATION_STRUCTURE_COMPACTED_SIZE, queryPool.get(), queryCount++);
+
+			cmdbufSubmitAndWait(cmdbufBlas, getComputeQueue(), 39);
+
+// TODO: SUBMIT THE BLAS BUILD FIRST AND AWAIT ITS COMPLETION WITH A SEMAPHORE SIGNAL
```
devshgraphicsprogramming, line 934: remove the comment, you're doing it

67_RayQueryGeometry/main.cpp
```diff
@@ -0,0 +1,1088 @@
...
+
+				// make sure instances upload complete first
+				{
+					SMemoryBarrier memBarrier;
+					memBarrier.srcStageMask = PIPELINE_STAGE_FLAGS::ALL_TRANSFER_BITS;
+					memBarrier.srcAccessMask = ACCESS_FLAGS::TRANSFER_WRITE_BIT;
+					memBarrier.dstStageMask = PIPELINE_STAGE_FLAGS::ACCELERATION_STRUCTURE_BUILD_BIT;
+					memBarrier.dstAccessMask = ACCESS_FLAGS::ACCELERATION_STRUCTURE_WRITE_BIT;
+					cmdbufTlas->pipelineBarrier(E_DEPENDENCY_FLAGS::EDF_NONE, { .memBarriers = {&memBarrier, 1} });
+				}
```
devshgraphicsprogramming, line 989: this also should contain and source flags for the BLAS compaction

67_RayQueryGeometry/app_resources/render.comp.hlsl
```diff
@@ -19,6 +19,16 @@ cbuffer CameraData
...
 
+// TODO: don't know if this is correct
+float3 getNormalsFromMask(int bn, int mask)
+{
+    float3 n = (float3)0;
+    n.x = float(asint((bn >> 24) & mask));
+    n.y = float((bn >> 16) & mask);
+    n.z = float((bn >> 8) & mask);
+    return n;
+}
```
devshgraphicsprogramming, line 30: what are you trying to do?

keptsecret: Read an int8_t3 from the buffer as an int32_t. So what I think needs doing is shift it down and read 8 bits as a float. I forgot to `asint` the other lines but don't think it's necessary

67_RayQueryGeometry/main.cpp
devshgraphicsprogramming: ok  `UNASSIGNED-CoreValidation-DrawState-QueryNotReset`

67_RayQueryGeometry/app_resources/render.comp.hlsl
devshgraphicsprogramming: read a `uint32_t` and then use the GLSL.450 extended intrinsic `unpack4x8Unorm`, implement it here: 
https://github.com/Devsh-Graphics-Programming/Nabla/blob/d37770f6dfab7f19de6c798ddd2c5a93256ee7ae/include/nbl/builtin/hlsl/spirv_intrinsics/core.hlsl#L198

67_RayQueryGeometry/main.cpp
```diff
@@ -0,0 +1,998 @@
...
+				memBarrier.dstStageMask = PIPELINE_STAGE_FLAGS::ACCELERATION_STRUCTURE_BUILD_BIT;
+				memBarrier.dstAccessMask = ACCESS_FLAGS::ACCELERATION_STRUCTURE_READ_BIT;
+				cmdbufBlas->pipelineBarrier(E_DEPENDENCY_FLAGS::EDF_NONE, { .memBarriers = {&memBarrier, 1} });
+			}
+
+			cmdbufBlas->resetQueryPool(queryPool.get(), 0, 1);
+
+			uint32_t queryCount = 0;
+			const IGPUAccelerationStructure* ases[1u] = { gpuBlas.get() };
+			cmdbufBlas->writeAccelerationStructureProperties({ ases, 1}, IQueryPool::ACCELERATION_STRUCTURE_COMPACTED_SIZE, queryPool.get(), queryCount++);
```
devshgraphicsprogramming, line 830: if the "queries need no barriers" thing is true, then you don't need a pipeline barrier.

67_RayQueryGeometry/app_resources/render.comp.hlsl
```diff
@@ -19,14 +21,12 @@ cbuffer CameraData
...
 {
-    float3 n = (float3)0;
-    n.x = float(asint((bn >> 24) & mask));
-    n.y = float((bn >> 16) & mask);
-    n.z = float((bn >> 8) & mask);
-    return n;
+    // this still doesn't feel right
+    // int8_t3 --> SSCALED (converts to float) --> unpackUnorm (read as uint --> float / 255)
+    float4 v = nbl::hlsl::spirv::unpackUnorm4x8(n) * 255.0;
+    return v.xyz;
```
devshgraphicsprogramming, line 29: the encoding scheme of the normal is kinda funny, you just need to normalize it instead of multiplying by any constant

devshgraphicsprogramming: there's also an `unpackSnorm4x8` and that will give you `[-127,127]` range

67_RayQueryGeometry/app_resources/render.comp.hlsl
```diff
@@ -100,6 +111,7 @@ void main(uint32_t3 threadID : SV_DispatchThreadID)
 
         float3 normalInterp = barycentrics.x * n0 + barycentrics.y * n1 + barycentrics.z * n2;
         color = float4(normalInterp, 1.0);
```
devshgraphicsprogramming: renormalize your normal after interpolation, leave the *0.5+0.5 to the very end

67_RayQueryGeometry/app_resources/render.comp.hlsl
```diff
@@ -98,20 +110,49 @@ void main(uint32_t3 threadID : SV_DispatchThreadID)
...
+        float3 n0, n1, n2;
+        switch (instID)
+        {
+            case OT_CUBE:
+            {
+                // this still doesn't feel right
+                // int8_t3 --> SSCALED (converts to float) --> unpackUnorm (read as uint --> float / 255)
+                n0 = (nbl::hlsl::spirv::unpackUnorm4x8(v0) * 255.0).yzw;
+                n1 = (nbl::hlsl::spirv::unpackUnorm4x8(v1) * 255.0).yzw;
+                n2 = (nbl::hlsl::spirv::unpackUnorm4x8(v2) * 255.0).yzw;
```
devshgraphicsprogramming, line 126: you might see that the *255.0 makes no difference to the normal value because of normalization

67_RayQueryGeometry/app_resources/render.comp.hlsl
```diff
@@ -98,20 +110,49 @@ void main(uint32_t3 threadID : SV_DispatchThreadID)
...
+                n0 = vk::RawBufferLoad<float3>(pc.geom[instID].vertexBufferAddress + i0 * vertexStride + byteOffset);
+                n1 = vk::RawBufferLoad<float3>(pc.geom[instID].vertexBufferAddress + i1 * vertexStride + byteOffset);
+                n2 = vk::RawBufferLoad<float3>(pc.geom[instID].vertexBufferAddress + i2 * vertexStride + byteOffset);
+            }
+        }
 
         float3 barycentrics = float3(0.0, query.CommittedTriangleBarycentrics());
         barycentrics.x = 1.0 - barycentrics.y - barycentrics.z;
 
         float3 normalInterp = barycentrics.x * n0 + barycentrics.y * n1 + barycentrics.z * n2;
```
devshgraphicsprogramming, line 153: the "compressed" normals (RGB8, RGB10) must be normalized before interpolation

67_RayQueryGeometry/app_resources/render.comp.hlsl
```diff
@@ -0,0 +1,159 @@
...
+
+    RayQuery<RAY_FLAG_FORCE_OPAQUE> query;
+    query.TraceRayInline(topLevelAS, 0, 0xFF, ray);
+
+    while (query.Proceed()) {}
+
+    if (query.CommittedStatus() == COMMITTED_TRIANGLE_HIT)
+    {
+        const int instID = query.CommittedInstanceIndex();
+        const int primID = query.CommittedPrimitiveIndex();
```
devshgraphicsprogramming, line 80: now time to do this in Inline SPIR-V and add the inline SPIR-V intrinsics to our `spirv_intrinsics`

67_RayQueryGeometry/app_resources/render.comp.hlsl
```diff
@@ -121,29 +121,30 @@ void main(uint32_t3 threadID : SV_DispatchThreadID)
             {
                 // this still doesn't feel right
                 // int8_t3 --> SSCALED (converts to float) --> unpackUnorm (read as uint --> float / 255)
-                n0 = (nbl::hlsl::spirv::unpackUnorm4x8(v0) * 255.0).yzw;
-                n1 = (nbl::hlsl::spirv::unpackUnorm4x8(v1) * 255.0).yzw;
-                n2 = (nbl::hlsl::spirv::unpackUnorm4x8(v2) * 255.0).yzw;
+                n0 = normalize(float3(asint(nbl::hlsl::spirv::unpackUnorm4x8(v0)).xyz));
+                n1 = normalize(float3(asint(nbl::hlsl::spirv::unpackUnorm4x8(v1)).xyz));
+                n2 = normalize(float3(asint(nbl::hlsl::spirv::unpackUnorm4x8(v2)).xyz));
+                n2 = any(isnan(n2)) ? float3(0, 0, 0) : n2; // sometimes n2 is (0, 0, 0), normalizing produces nan, probably reading wrong somehow
```
devshgraphicsprogramming, line 127: thats concerning, a cube's normal should never be 0,0,0

67_RayQueryGeometry/app_resources/render.comp.hlsl
```diff
@@ -0,0 +1,160 @@
...
+        const uint byteOffset = byteOffsets[instID];
+
+        uint i0, i1, i2;
+        switch (indexType)
+        {
+            case 0: // EIT_16BIT
+            {
+                i0 = uint32_t(vk::RawBufferLoad<uint16_t>(pc.geom[instID].indexBufferAddress + (idxOffset + 0) * sizeof(uint16_t)));
+                i1 = uint32_t(vk::RawBufferLoad<uint16_t>(pc.geom[instID].indexBufferAddress + (idxOffset + 1) * sizeof(uint16_t)));
+                i2 = uint32_t(vk::RawBufferLoad<uint16_t>(pc.geom[instID].indexBufferAddress + (idxOffset + 2) * sizeof(uint16_t)));
```
devshgraphicsprogramming, line 95: set the alignment on the `vk::RawBufferLoad`

67_RayQueryGeometry/common.hlsl
devshgraphicsprogramming: still a TODO

67_RayQueryGeometry/main.cpp
devshgraphicsprogramming: delete the renderpass, still a TODO

67_RayQueryGeometry/main.cpp
devshgraphicsprogramming: still a TODO

67_RayQueryGeometry/main.cpp
devshgraphicsprogramming: still a TODO

67_RayQueryGeometry/main.cpp
devshgraphicsprogramming: still a TODO

67_RayQueryGeometry/main.cpp
```diff
@@ -0,0 +1,1096 @@
...
+			memcpy(uboData.invMVP, invModelViewProjectionMatrix.pointer(), sizeof(uboData.invMVP));
+			memcpy(uboData.V, viewMatrix.pointer(), sizeof(uboData.V));
+			memcpy(uboData.P, projectionMatrix.pointer(), sizeof(uboData.P));
+			{
+				SBufferRange<IGPUBuffer> range;
+				range.buffer = smart_refctd_ptr(cameraBuffer);
+				range.size = cameraBuffer->getSize();
+
+				cmdbuf->updateBuffer(range, &uboData);
+			}
```
devshgraphicsprogramming, line 380: 1. get rid of the UBO
2. your update was wrong because you didn't have a pipeline barrier between the update the the compute shader, I'm surprised validation layers didn't scream at you

67_RayQueryGeometry/main.cpp
devshgraphicsprogramming: done?

67_RayQueryGeometry/main.cpp
```diff
@@ -0,0 +1,1096 @@
...
+
+					IGPUBottomLevelAccelerationStructure::BuildRangeInfo buildRangeInfos[1u];
+					buildRangeInfos[0].primitiveCount = primitiveCounts[i];
+					buildRangeInfos[0].primitiveByteOffset = 0u;
+					buildRangeInfos[0].firstVertex = 0u;
+					buildRangeInfos[0].transformByteOffset = 0u;
+					IGPUBottomLevelAccelerationStructure::BuildRangeInfo* pRangeInfos[1u];
+					pRangeInfos[0] = &buildRangeInfos[0];
+
+					cmdbufBlas->buildAccelerationStructures({ &blasBuildInfos[i], 1 }, pRangeInfos);
```
devshgraphicsprogramming, line 902: why areyou building one and barriering one at a time?

Build all blas at once! that way you can hope to saturate the GPU (the builds run in parallel)

67_RayQueryGeometry/main.cpp
```diff
@@ -0,0 +1,1096 @@
...
+						memBarrier.srcStageMask = PIPELINE_STAGE_FLAGS::ACCELERATION_STRUCTURE_BUILD_BIT;
+						memBarrier.srcAccessMask = ACCESS_FLAGS::ACCELERATION_STRUCTURE_WRITE_BIT;
+						memBarrier.dstStageMask = PIPELINE_STAGE_FLAGS::ACCELERATION_STRUCTURE_BUILD_BIT;
+						memBarrier.dstAccessMask = ACCESS_FLAGS::ACCELERATION_STRUCTURE_READ_BIT;
+						cmdbufBlas->pipelineBarrier(E_DEPENDENCY_FLAGS::EDF_NONE, { .memBarriers = {&memBarrier, 1} });
+					}
+
+					const IGPUAccelerationStructure* ases[1u] = { gpuBlas[i].get() };
+					cmdbufBlas->writeAccelerationStructureProperties({ ases, 1 }, IQueryPool::ACCELERATION_STRUCTURE_COMPACTED_SIZE,
+						queryPool.get(), queryCount++);
```
devshgraphicsprogramming, line 915: get all queries at once, at the end

67_RayQueryGeometry/main.cpp
devshgraphicsprogramming: hmm technically BLAS compaction counts as a TRANSFER because its a copy. so flags all good

67_RayQueryGeometry/main.cpp
devshgraphicsprogramming: still TODO

67_RayQueryGeometry/app_resources/render.comp.hlsl
```diff
@@ -0,0 +1,160 @@
+#include "common.hlsl"
+
+#include "nbl/builtin/hlsl/spirv_intrinsics/core.hlsl"
+
+struct SPushConstants
+{
+    SGeomBDA geom[OT_COUNT];
+};
```
devshgraphicsprogramming, line 8: fun fact, push constants CANNOT BE DYNAMICALLY INDEXED (@kpentaris got cut on it once)

You want to have the `SGeomBDA` in a IGPUBuffer somewhere (maybe call it `SGeomInfo), also put the vertex offsets and strides in there or at least `SVertexInfo`.

then you want a BDA address to the `SGeomBDA` in the push constants and `vk::RawBufferLoad` them

devshgraphicsprogramming: and the camera data you want in the push constants

67_RayQueryGeometry/render.comp.hlsl
devshgraphicsprogramming: still a TODO

67_RayQueryGeometry/app_resources/render.comp.hlsl
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// How the normals are packed
+// cube                     uint32_t - 8, 8, 8, 8
+// sphere, cylinder, arrow  uint32_t - 10, 10, 10, 2
+// rectangle, disk          float3
+// cone                     uint32_t - 10, 10, 10, 2
+// icosphere                float3
+static const uint indexTypes[OT_COUNT]      = { 0, 1, 0, 0, 2, 0, 0, 1 };
+static const uint vertexStrides[OT_COUNT]   = { 24, 28, 28, 32, 32, 28, 20, 32 };
```
devshgraphicsprogramming, line 41: everything should probably live together in one struct, vertex buffer offset, index buffer offset, geometry type, index  type, vertex strides

67_RayQueryGeometry/app_resources/render.comp.hlsl
```diff
@@ -0,0 +1,160 @@
...
+
+float3 unpackNormals3x10(uint32_t v)
+{
+    // host side changes float32_t3 to EF_A2B10G10R10_SNORM_PACK32
+    float3 n = (float3)0;
+    const uint mask = 0x3FF;    // 10 bits
+    n.x = float(asint((v >> 20) & mask));
+    n.y = float(asint((v >> 10) & mask));
+    n.z = float(asint((v >> 0) & mask));
+    return n;
```
devshgraphicsprogramming, line 31: normals are always snorm and not unorm, snorm unpacks differently to unrom (its not a simple *0.5-0.5 from a unorm)

67_RayQueryGeometry/app_resources/render.comp.hlsl
devshgraphicsprogramming: still a todo

67_RayQueryGeometry/app_resources/render.comp.hlsl
devshgraphicsprogramming: still a todo

67_RayQueryGeometry/main.cpp
keptsecret: Looks like a blit is used as well in example 27

67_RayQueryGeometry/main.cpp
```diff
@@ -158,56 +106,33 @@ class RayQueryGeometryApp final : public examples::SimpleWindowedApplication, pu
...
+					.format = asset::EF_R16G16B16A16_SFLOAT,
+					.extent = {WIN_W, WIN_H, 1},
+					.mipLevels = 1,
+					.arrayLayers = 1,
+					.flags = IImage::ECF_NONE,
+					.usage = core::bitflag(asset::IImage::EUF_STORAGE_BIT) | asset::IImage::EUF_TRANSFER_SRC_BIT
+				}
+			});
+			if (!outHDRImage || !m_device->allocate(outHDRImage->getMemoryReqs(), outHDRImage.get()).isValid())
+				return logFail("Could not create HDR Image");
```
devshgraphicsprogramming, line 122: you don't need a HDR image for this example, you can probably store straight to swapchain image if you want

67_RayQueryGeometry/main.cpp
devshgraphicsprogramming: ah yes, indeed probably because the example is meant to do simple Whitted Raytracing with accumulation and most desktops can't be relied upon to have a HDR swapchain

67_RayQueryGeometry/app_resources/common.hlsl
```diff
@@ -0,0 +1,46 @@
...
+
+NBL_CONSTEXPR uint32_t WorkgroupSize = 16;
+
+struct SGeomInfo
+{
+    uint64_t vertexBufferAddress;
+    uint64_t indexBufferAddress;
+
+	uint32_t indexType; // 16 bit, 32 bit or none
+	uint32_t vertexStride;
```
devshgraphicsprogramming, line 14: pro tip, your struct is already 64 byte aligned, you may wanna keep the other members down to have no padding, so lets say
```
uint32_t indexType : 2;
uint32_t vertexStride : 30;
uint32_t byteOffset;
```
which add up to 64 bits

67_RayQueryGeometry/main.cpp
```diff
@@ -0,0 +1,1043 @@
...
+					return logFail("Failed to build BLAS");
+
+				{
+					SMemoryBarrier memBarrier;
+					memBarrier.srcStageMask = PIPELINE_STAGE_FLAGS::ACCELERATION_STRUCTURE_BUILD_BIT;
+					memBarrier.srcAccessMask = ACCESS_FLAGS::ACCELERATION_STRUCTURE_WRITE_BIT;
+					memBarrier.dstStageMask = PIPELINE_STAGE_FLAGS::ACCELERATION_STRUCTURE_BUILD_BIT;
+					memBarrier.dstAccessMask = ACCESS_FLAGS::ACCELERATION_STRUCTURE_READ_BIT;
+					cmdbufBlas->pipelineBarrier(E_DEPENDENCY_FLAGS::EDF_NONE, { .memBarriers = {&memBarrier, 1} });
+				}
```
devshgraphicsprogramming, line 848: id you submit and wait for the command buffer, you don't need to do a pipeline barrier, the semaphore already does it

PR #165, opened 2024-12-30
BxDF tests

Przemog1:


devshgraphicsprogramming:
why do you have random things from sorakrit here instead of your own branch against his or `master` ?

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -0,0 +1,455 @@
...
+inline float rngFloat01(NBL_REF_ARG(nbl::hlsl::Xoroshiro64Star) rng)
+{
+    return (float)rng() / numeric_limits<uint32_t>::max;
+}
+
+inline float32_t3 rngFloat301(NBL_REF_ARG(nbl::hlsl::Xoroshiro64Star) rng)
+{
+    return float32_t3(rngFloat01(rng), rngFloat01(rng), rngFloat01(rng));
+}
+
```
devshgraphicsprogramming, line 33: I'd make a `rngUniformDist<T>` where `T` can be `float32_t` or `vector<float_t,N>` , suffix in the name like 301 is weird to read

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -0,0 +1,455 @@
...
+{
+    static SBxDFTestResources create(uint32_t2 seed)
+    {
+        SBxDFTestResources retval;
+
+        retval.rng = nbl::hlsl::Xoroshiro64Star::construct(seed);
+        retval.u = float32_t3(rngFloat01(retval.rng), rngFloat01(retval.rng), 0.0);
+
+        retval.V.direction = nbl::hlsl::normalize<float32_t3>(rngFloat301(retval.rng));
+        retval.N = nbl::hlsl::normalize<float32_t3>(rngFloat301(retval.rng));
```
devshgraphicsprogramming, line 45: directions should be generated differently, there's a [0,1]^2 -> Hemisphere / Sphere transform code you should test first and then use throughout

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -0,0 +1,455 @@
...
+        retval.T = nbl::hlsl::normalize<float32_t3>(rngFloat301(retval.rng));
+
+        retval.T = nbl::hlsl::normalize<float32_t3>(retval.T - nbl::hlsl::dot<float32_t3>(retval.T, retval.N) * retval.N); // gram schmidt
```
devshgraphicsprogramming, line 48: you can still generate `T` almost colinear to `N` and end up with numerical issues, personally I'd use the frisvad basis function + random rotation around N (only one random number needed)

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -0,0 +1,455 @@
...
+
+
+struct SBxDFTestResources
+{
+    static SBxDFTestResources create(uint32_t2 seed)
+    {
+        SBxDFTestResources retval;
+
+        retval.rng = nbl::hlsl::Xoroshiro64Star::construct(seed);
+        retval.u = float32_t3(rngFloat01(retval.rng), rngFloat01(retval.rng), 0.0);
```
devshgraphicsprogramming, line 42: the third random number is sometimes necessary to discern discrete BxDF branches:
- reflection vs refraction in e.g. glass
- diffuse vs specular in e.g. plastic
- choosing a leaf BxDF in a mixture/tree of BxDFs

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -0,0 +1,455 @@
...
+    ray_dir_info_t dV(int axis)
+    {
+        float32_t3 d = (float32_t3)0.0;
+        d[axis] += h;
+        ray_dir_info_t retval;
+        retval.direction = V.direction + d;
+        return retval;
+    }
+
+    float h = 0.001;
```
devshgraphicsprogramming, line 69: 1. what do you need `dV` for?
2. this is not the way to vary `V` parameter

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -0,0 +1,455 @@
...
+};
+
+struct STestMeta
+{
+    float32_t4 result;
+#ifndef __HLSL_VERSION
+    std::string bxdfName;
+    std::string testName;
+#endif
+};
```
devshgraphicsprogramming, line 95: returning a `result` as a vec4 is quite a bit meh, I'd actually provide a callback (via static polymorphism) which is only called upon a failure
```cpp
struct FailureCallbackConceptExample
{
   void __call(NBL_CONST_REF_ARG(SBxDFTestResources) failedFor, NBL_CONST_REF_ARG(sample_t) failedAt);
};
```


then you can print in C++, mark booleans, append to vector, break & repeat call or whatever

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -78,6 +80,9 @@ struct SBxDFTestResources
     float32_t2 alpha;
     float eta;
     float32_t3x2 ior;
+
+    float32_t3 eta2;    // what is this?
```
devshgraphicsprogramming, line 84: `eta` squared, compute on the fly, don't leave stuff like this here

66_HLSLBxDFTests/app_resources/tests.hlsl
devshgraphicsprogramming: perfect C++ callback
```cpp
printf(...your error message...);
for (volatile bool repeat=true; gDebuggerAttached&&repeat; )
{
   repeat = false;
   __debugbreak();
   failedFor.test(failedAt);
}
```

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -0,0 +1,455 @@
...
+        t.init(seed);
+        if NBL_CONSTEXPR_FUNC (is_microfacet_brdf_v<BxDF> || is_microfacet_bsdf_v<BxDF>)
+            t.template initBxDF<aniso>(t.rc);
+        else
+            t.initBxDF(t.rc);
+
+        t.meta.result = t.test();
+        t.meta.testName = "u offset";
+        return t.meta;
+    }
```
devshgraphicsprogramming, line 391: setting those thousands of times will be expensive, function should be `void` and not return anything, callback instead of meta

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -0,0 +1,455 @@
...
+        float32_t2x2 m = float32_t2x2(sx.TdotL - s.TdotL, sy.TdotL - s.TdotL, sx.BdotL - s.BdotL, sy.BdotL - s.BdotL);
+        float det = nbl::hlsl::determinant<float32_t2x2>(m);
+
+        return float32_t4(nbl::hlsl::abs<float32_t3>(pdf.value() - brdf), nbl::hlsl::abs<float>(det*pdf.pdf/s.NdotL) * 0.5);
+    }
+
+    static STestMeta run(uint32_t2 seed)
+    {
+        this_t t;
+        t.init(seed);
```
devshgraphicsprogramming, line 382: I'd make the function take `uint32_t` as an argument, then run a hash function (wide not deep)
https://rene.ruhr/gfx/gpuhash/

This will make it easy to invoke multiple `run(uint32_t invocationID)` in parallel independently, be it on CPU with `for` loop or GPU with `gl_GlobalInvocationID.x`

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -0,0 +1,455 @@
...
+        quotient_pdf_t pdf;
+        float32_t3 brdf;
+        aniso_cache cache, dummy;
+
+        iso_interaction isointerx = iso_interaction::create(base_t::rc.dV(0), base_t::rc.N);
+        aniso_interaction anisointerx = aniso_interaction::create(isointerx, base_t::rc.T, base_t::rc.B);
+        iso_interaction isointery = iso_interaction::create(base_t::rc.dV(1), base_t::rc.N);
+        aniso_interaction anisointery = aniso_interaction::create(isointery, base_t::rc.T, base_t::rc.B);
+        iso_interaction isointerz = iso_interaction::create(base_t::rc.dV(2), base_t::rc.N);
+        aniso_interaction anisointerz = aniso_interaction::create(isointerz, base_t::rc.T, base_t::rc.B);
```
devshgraphicsprogramming, line 413: there's no point to this test existing, the test I've outlined ONLY makes sense for varying (differentiating) the 2D random number `xi`, because thats the uniform distribution that must be stretched and squeezed to form the PDF of the importance sampler

66_HLSLBxDFTests/main.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include <iostream>
+#include <iomanip>
+
+#include <nbl/builtin/hlsl/cpp_compat.hlsl>
+
+using namespace nbl::hlsl;
+
+#include "app_resources/tests.hlsl"
+
+#define ASSERT_ZERO(x) (assert(all<bool32_t4>((x.result) < float32_t4(1e-3))));
```
devshgraphicsprogramming, line 11: use a nice callback

66_HLSLBxDFTests/main.cpp
```diff
@@ -0,0 +1,59 @@
...
+    printResult(TestUOffset<bxdf::reflection::SBeckmannBxDF<sample_t, iso_cache, aniso_cache>,false>::run(state));
+    ASSERT_ZERO((TestUOffset<bxdf::reflection::SBeckmannBxDF<sample_t, iso_cache, aniso_cache>,true>::run(state)));
+    printResult(TestUOffset<bxdf::reflection::SGGXBxDF<sample_t, iso_cache, aniso_cache>,false>::run(state));
+    ASSERT_ZERO((TestUOffset<bxdf::reflection::SGGXBxDF<sample_t, iso_cache, aniso_cache>,true>::run(state)));
+
+    ASSERT_ZERO((TestUOffset<bxdf::transmission::SLambertianBxDF<sample_t, iso_interaction, aniso_interaction>>::run(state)));
+    printResult(TestUOffset<bxdf::transmission::SBeckmannDielectricBxDF<sample_t, iso_cache, aniso_cache>,false>::run(state));
+    ASSERT_ZERO((TestUOffset<bxdf::transmission::SBeckmannDielectricBxDF<sample_t, iso_cache, aniso_cache>,true>::run(state)));
+    printResult(TestUOffset<bxdf::transmission::SGGXDielectricBxDF<sample_t, iso_cache, aniso_cache>,false>::run(state));
+    printResult(TestUOffset<bxdf::transmission::SGGXDielectricBxDF<sample_t, iso_cache, aniso_cache>,true>::run(state));
```
devshgraphicsprogramming: you need to run the tests thousands of times with different seeds, I'd use `for_each` over a `ranges::iota_view` with `par_unseq` policy to perform these tests on all CPU cores at once.
https://godbolt.org/z/xvxE7jzE3

this connects with https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/165/files#r1906914911

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -0,0 +1,455 @@
...
+            sy = base_t::bxdf.generate(base_t::anisointer, base_t::rc.u + float32_t3(0,base_t::rc.h,0));
+        }
+        if NBL_CONSTEXPR_FUNC (is_microfacet_bsdf_v<BxDF>)
+        {
+            s = base_t::bxdf.generate(base_t::anisointer, base_t::rc.u, cache);
+            float32_t3 ux = base_t::rc.u + float32_t3(base_t::rc.h,0,0);
+            sx = base_t::bxdf.generate(base_t::anisointer, ux, dummy);
+            float32_t3 uy = base_t::rc.u + float32_t3(0,base_t::rc.h,0);
+            sy = base_t::bxdf.generate(base_t::anisointer, uy, dummy);
+        }
```
devshgraphicsprogramming: `h` should be a parameter of each individual specialized BxDF test, because its specific to the differential test and also may need to change depending on the BxDF getting tested

devshgraphicsprogramming: also maybe compute the perturbed `u` above for readability

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -0,0 +1,455 @@
...
+                pdf = base_t::bxdf.quotient_and_pdf(s, base_t::isointer, isocache);
+                brdf = float32_t3(base_t::bxdf.eval(s, base_t::isointer, isocache));
+            }
+        }
+
+        // get jacobian
+        float32_t2x2 m = float32_t2x2(sx.TdotL - s.TdotL, sy.TdotL - s.TdotL, sx.BdotL - s.BdotL, sy.BdotL - s.BdotL);
+        float det = nbl::hlsl::determinant<float32_t2x2>(m);
+
+        return float32_t4(nbl::hlsl::abs<float32_t3>(pdf.value() - brdf), nbl::hlsl::abs<float>(det*pdf.pdf/s.NdotL) * 0.5);
```
devshgraphicsprogramming, line 376: this needs to be done better semantically with a callback, one thing for each individually.

Alos `pdf.value()` does not equal `brdf` in general!!!!!

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -0,0 +1,455 @@
...
+                brdf = float32_t3(base_t::bxdf.eval(s, base_t::anisointer, cache));
+            }
+            else
+            {
+                iso_cache isocache = (iso_cache)cache;
+                pdf = base_t::bxdf.quotient_and_pdf(s, base_t::isointer, isocache);
+                brdf = float32_t3(base_t::bxdf.eval(s, base_t::isointer, isocache));
+            }
+        }
+
```
devshgraphicsprogramming, line 371: you need to skip checking anything that produces an "impossible" `s`, so if `eval` or `quotient` is 0, you can skip the rest of the function (but not before you check pdf and quotient for 0 and INF respectively - the checks also check for NaN)

add more simple tests for:
1. `pdf>0` because something you generated cannot have 0 probability of getting generated
2. `quotient<INF` always because our importance sampler's job is to prevent that!
3. positivity, all `pdf`, `quotient`, and `eval` need to be `>=0`
4. recprocity (eval must be equal if we swap L and V around, make a method/function to swap L and V around in `s`, the interaction, the `bxdf` itself and the `cache`)

Leave lots of comments.

devshgraphicsprogramming: Also when `pdf` is close to `INF` (above some super high threshold) you need to back off the last test you have here

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -0,0 +1,455 @@
...
+        retval.T = nbl::hlsl::normalize<float32_t3>(retval.T - nbl::hlsl::dot<float32_t3>(retval.T, retval.N) * retval.N); // gram schmidt
+        retval.B = nbl::hlsl::cross<float32_t3>(retval.N, retval.T);
+
+        retval.alpha.x = rngFloat01(retval.rng);
+        retval.alpha.y = rngFloat01(retval.rng);
+        retval.eta = 1.3;// rngFloat01(retval.rng) + 1.0;
+        retval.ior = float32_t3x2(1.02, 1.0,      // randomize at some point?
+                                1.3, 2.0,
+                                1.02, 1.0);
```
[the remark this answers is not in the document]
keptsecret: Do we want to keep ior as a matrix in the BxDF?
For the BRDFs https://github.com/Devsh-Graphics-Programming/Nabla/blob/6fc78a46f466d176f1b3a4c84f18f7448d3cd894/include/nbl/builtin/hlsl/bxdf/reflection.hlsl#L361

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -39,21 +74,23 @@ struct SBxDFTestResources
...
-
-        retval.V.direction = nbl::hlsl::normalize<float32_t3>(rngFloat301(retval.rng));
-        retval.N = nbl::hlsl::normalize<float32_t3>(rngFloat301(retval.rng));
-        retval.T = nbl::hlsl::normalize<float32_t3>(rngFloat301(retval.rng));
+        retval.u = float32_t3(rngUniformDist<float32_t2>(retval.rng), 0.0);
 
-        retval.T = nbl::hlsl::normalize<float32_t3>(retval.T - nbl::hlsl::dot<float32_t3>(retval.T, retval.N) * retval.N); // gram schmidt
-        retval.B = nbl::hlsl::cross<float32_t3>(retval.N, retval.T);
+        retval.V.direction = nbl::hlsl::normalize<float32_t3>(projected_hemisphere_generate<float>(rngUniformDist<float32_t2>(retval.rng)));
+        retval.N = nbl::hlsl::normalize<float32_t3>(projected_hemisphere_generate<float>(rngUniformDist<float32_t2>(retval.rng)));
```
devshgraphicsprogramming, line 80: not projected hemisphere, just hemisphere.

Also not hemisphere but full sphere, because its good to test that BRDFs handle invalid (below hemisphere) correctly

66_HLSLBxDFTests/app_resources/tests.hlsl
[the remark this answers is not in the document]
devshgraphicsprogramming: no dont keep it was a super bad idea, I also want to make everything templated so that `vector<T,3>` isn't implied for the colours and we could do other number of spectral buckets (1, 2 or 4 for spectral rendering)

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -22,6 +22,18 @@ using iso_cache = bxdf::SIsotropicMicrofacetCache<float>;
 using aniso_cache = bxdf::SAnisotropicMicrofacetCache<float>;
 using quotient_pdf_t = bxdf::quotient_and_pdf<float32_t3, float>;
 
+uint32_t pcg_hash(uint32_t v)
+{
+    uint32_t state = v * 747796405u + 2891336453u;
+	uint32_t word = ((state >> ((state >> 28u) + 4u)) ^ state) * 277803737u;
+	return (word >> 22u) ^ word;
+}
```
devshgraphicsprogramming, line 34: this is useful in its own header like `xoroshiro` is

62_CAD/shaders/main_pipeline/vertex_shader.hlsl
```diff
@@ -87,8 +87,10 @@ void dilateHatch<false>(out float2 outOffsetVec, out float2 outUV, const float2
 
 PSInput main(uint vertexID : SV_VertexID)
 {
-    float a = 0.0f;
-    float b = hlsl::cos(a);
+    // TODO[Przemek]: Disable Everything here and do your own thing as we already discussed, but let's have the same PSInput data passed to fragment.
+    // your programmable pulling will use the baseVertexBufferAddress BDA address and `vertexID` to RawBufferLoad it's vertex. 
+    // ~~Later, most likely We will require pulling all 3 vertices of the triangle, that's where you need to know which triangle you're currently on, and instead of objectID = vertexID/4 which we currently do, you will do vertexID/3 and pull all 3 of it's vertices.~~
+    // Ok, brainfart, a vertex can belong to multiple triangles, I was thinking of AA but triangles share vertices, nevermind my comment above.
```
devshgraphicsprogramming: why is this on the `bxdf_unit_tests` brach ?

@Erfan-Ahmadi @Przemog1 ?

devshgraphicsprogramming:
@keptsecret where's the branch with the HLSL and GLSL path tracer changes?

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -36,6 +35,18 @@ namespace IMATH = Imath;
...
+template<>
+struct std::hash<float32_t2>
+{
+    size_t operator()(const float32_t2& v) const noexcept
+    {
+        size_t v1 = std::hash<float32_t>()(v.x);
+        size_t v2 = std::hash<float32_t>()(v.y);
+        return v1 ^ (v2 << 1);
+    }
+};
```
devshgraphicsprogramming, line 49: make one in `cpp_compat` under `__HLSL_VERSION` note that we have our own `hash_combine` which is very similar to GLM's https://github.com/g-truc/glm/blob/master/glm/gtx/hash.inl

You can fairly easily make `hash_helper` which you can partial specialize for vectors and matrices of all types and dimensions

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -82,6 +105,46 @@ T rngUniformDist(NBL_REF_ARG(nbl::hlsl::Xoroshiro64Star) rng)
...
+template<typename T>
+bool checkLt(T a, T b)
+{
+    return nbl::hlsl::all<vector<bool, vector_traits<T>::Dimension> >(a < b);
+}
+
+template<typename T>
+bool checkZero(T a, float32_t eps)
+{
+    return nbl::hlsl::all<vector<bool, vector_traits<T>::Dimension> >(nbl::hlsl::abs<T>(a) < (T)eps);
```
devshgraphicsprogramming, line 123: use promote not a C-cast check it everywhere

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -82,6 +105,46 @@ T rngUniformDist(NBL_REF_ARG(nbl::hlsl::Xoroshiro64Star) rng)
...
+{
+    return float32_t3(std::cos(theta_phi.y) * std::cos(theta_phi.x),
+                        std::sin(theta_phi.y) * std::cos(theta_phi.x),
+                        std::sin(theta_phi.x));
+}
+
+inline float32_t2 cartesianToPolar(float32_t3 coords)
+{
+    return float32_t2(std::acos(clamp<float>(coords.z, -1, 1)), std::atan2(coords.y, coords.x));
+}
```
devshgraphicsprogramming, line 146: make these into general utilities in `nbl/builtin/hlsl/math`

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -91,18 +154,21 @@ struct SBxDFTestResources
...
+        retval.V.direction = nbl::hlsl::normalize<float32_t3>(sampling::UniformSphere<float>::generate(rngUniformDist<float32_t2>(retval.rng)));
+        retval.N = nbl::hlsl::normalize<float32_t3>(sampling::UniformSphere<float>::generate(rngUniformDist<float32_t2>(retval.rng)));
         
-        float32_t2x3 tb = math::frisvad<float>(retval.N);
+        float32_t3 tangent, bitangent;
+        math::frisvad<float32_t3>(retval.N, tangent, bitangent);
+        tangent = nbl::hlsl::normalize<float32_t3>(tangent);
+        bitangent = nbl::hlsl::normalize<float32_t3>(bitangent);
 #ifndef __HLSL_VERSION
         const float angle = 2 * numbers::pi<float> * rngUniformDist<float>(retval.rng);
```
devshgraphicsprogramming, line 165: use 2.f instead of 2 as the literal

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -113,8 +179,8 @@ struct SBxDFTestResources
         return retval;
     }
 
-    // epsilon
-    float eps = 1e-3;
+    float eps = 1e-3;   // epsilon
+    uint32_t state;     // init state seed, for debugging
```
devshgraphicsprogramming, line 183: isn't the seed for xoroshiro 64bit ?

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -338,21 +419,21 @@ struct is_basic_brdf : bool_constant<
...
 
 template<class T>
 struct is_microfacet_bsdf : bool_constant<
-    is_same<T, bxdf::transmission::SBeckmannDielectricBxDF<sample_t, iso_cache, aniso_cache, spectral_t>>::value ||
-    is_same<T, bxdf::transmission::SGGXDielectricBxDF<sample_t, iso_cache, aniso_cache, spectral_t>>::value
+    is_same<T, bxdf::transmission::SBeckmannDielectricBxDF<sample_t, iso_interaction, aniso_interaction, iso_cache, aniso_cache, spectral_t>>::value ||
+    is_same<T, bxdf::transmission::SGGXDielectricBxDF<sample_t, iso_interaction, aniso_interaction, iso_cache, aniso_cache, spectral_t>>::value
 > {};
 
 template<class T>
```
devshgraphicsprogramming, line 439: there should be a general trait class in the `nbl/builtin/hlsl/bxdf` which tells us these things

btw what do you mean by basic ? I think you're checking just NOT microfacet.

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -91,18 +154,21 @@ struct SBxDFTestResources
         retval.rng = nbl::hlsl::Xoroshiro64Star::construct(seed);
         retval.u = float32_t3(rngUniformDist<float32_t2>(retval.rng), 0.0);
```
devshgraphicsprogramming, line 155: https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/165#discussion_r1906863399

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -113,8 +179,8 @@ struct SBxDFTestResources
         return retval;
```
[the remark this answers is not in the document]
devshgraphicsprogramming: P.s. take the luma coefficients from our cIE XYZ colorpsace headers

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -129,15 +195,29 @@ struct SBxDFTestResources
     float32_t3 luma_coeff;
 };
 
+struct STestInitParams
+{
+    bool logInfo;
+    uint32_t state;
```
devshgraphicsprogramming, line 201: what is `state` ?

66_HLSLBxDFTests/app_resources/tests.hlsl
```diff
@@ -91,18 +154,21 @@ struct SBxDFTestResources
...
-        retval.T = rot * tb[0];
-        retval.B = rot * tb[1];
+        retval.T = rot * tangent;
+        retval.B = rot * bitangent;
 #else
-        retval.T = tb[0];
-        retval.B = tb[1];
+        retval.T = tangent;
+        retval.B = bitangent;
 #endif
```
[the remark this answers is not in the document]
keptsecret: I've changed it to use the `rotation_mat` in `math/linalg/transform.hlsl` instead. Do we still want this quaternion.hlsl?

keptsecret:
Superseded by #236

PR #169, opened 2025-01-10
initialize blur example

alichraghi:


26_Blur/app_resources/common.hlsl
```diff
@@ -2,7 +2,18 @@
...
 static const uint16_t WORKGROUP_SIZE = 256;
 
+enum EdgeWrapMode : uint32_t {
+	WRAP_MODE_CLAMP_TO_BORDER,
+	WRAP_MODE_CLAMP_TO_EDGE,
+	WRAP_MODE_REPEAT,
+	WRAP_MODE_MIRROR,
+
+	WRAP_MODE_MAX,
+};
```
devshgraphicsprogramming, line 12: why not move the `ISampler` wrap modes to HLSL nad probide a backward alias ?

26_Blur/app_resources/shader.comp.hlsl
```diff
@@ -0,0 +1,221 @@
...
+                result -= repeatedValue;
+            } break;
+            case WRAP_MODE_MIRROR:
+            {
+                const uint16_t mirroredIdx = (uint16_t)((-leftIdx / width) % 2 == 0 ? (-leftIdx) % width : lastIdx - ((-leftIdx) % width));
+                float32_t mirrored;
+                smemAccessor.get(mirroredIdx, mirrored);
+                result -= mirrored;
+            } break;
+        }
```
devshgraphicsprogramming, line 126: why do you couple wrap offset calculation with fetching memory?

26_Blur/app_resources/common.hlsl
```diff
@@ -0,0 +1,19 @@
...
+
+struct PushConstants
+{
+    uint32_t flip;
+    uint32_t radius;
+    uint32_t edgeWrapMode;
```
devshgraphicsprogramming, line 18: bitfields

26_Blur/app_resources/shader.comp.hlsl
alichraghi: this part is WIP and mostly broken but also note that `last`, `leftFloor`, etc aren't indices

26_Blur/app_resources/common.hlsl
devshgraphicsprogramming: still not done

PR #172, opened 2025-01-22
New tgmath

Przemog1:


devshgraphicsprogramming:
@keptsecret can you merge `master` into `bxdf_unit_tests` so there's less of a diff here?

22_CppCompat/app_resources/tgmathTest.comp.hlsl
```diff
@@ -0,0 +1,16 @@
...
+#include "common.hlsl"
+
+[[vk::binding(0, 0)]] RWStructuredBuffer<TgmathIntputTestValues> inputTestValues;
+[[vk::binding(1, 0)]] RWStructuredBuffer<TgmathTestValues> outputTestValues;
+
+[numthreads(16, 1, 1)]
+void main(uint3 invocationID : SV_DispatchThreadID)
+{
+    if(invocationID.x == 0)
+        outputTestValues[0].fillTestValues(inputTestValues[0]);
```
devshgraphicsprogramming, line 15: you could try more values than just 1 set

22_CppCompat/app_resources/common.hlsl
```diff
@@ -25,11 +28,352 @@
...
+	float ldexp;
+
+	float32_t3 floorVec;
+#ifndef __HLSL_VERSION
+	nbl::hlsl::vector<int, 3> isnanVec;
+	nbl::hlsl::vector<int, 3> isinfVec;
+#else
+	vector<int, 3> isnanVec;
+	vector<int, 3> isinfVec;
+#endif
```
devshgraphicsprogramming, line 132: leave a comment that we can't fix this because `using namespace nbl::hlsl` would cause ambiguous math functions below and we can't add a `nbl::hlsl` alias for the builtin hLSL vector type because of https://github.com/microsoft/DirectXShaderCompiler/issues/7035

22_CppCompat/Testers.h
```diff
@@ -0,0 +1,863 @@
...
+        m_logger->log(msg, system::ILogger::ELL_ERROR, std::forward<Args>(args)...);
+        exit(-1);
+    }
+};
+
+class CTgmathTester final : public ITester
+{
+public:
+    void performTests()
+    {
```
devshgraphicsprogramming, line 339: why aren't the testers in their own includes?

22_CppCompat/app_resources/tgmathTest.comp.hlsl
devshgraphicsprogramming: there's something I don't get, you do 16 dispatches of 16 wide workgroups (should be at least 64 workgroup size if not 256 btw), so you have 16*16=256 invocations and I can see you in the Tester init code actually use random distributions to provide multiple test inputs.

Then why is only the first invocation of every workgroup (there are 16) only reading from and writing to the first global memory location?

Przemog1: corrected, now it is a single dispatch of 256 wide workgroup where only the first invocation does anything. I wanted this test to be convenient not efficient

PR #179, opened 2025-01-27
CAD: multi entrypoint shader

alichraghi:
Depends on https://github.com/Devsh-Graphics-Programming/Nabla/pull/754


devshgraphicsprogramming:
and this works, correct?

62_CAD/main.cpp
```diff
@@ -860,14 +860,9 @@ class ComputerAidedDesign final : public examples::SimpleWindowedApplication, pu
...
-		// Main Pipeline Shaders
-		std::array<smart_refctd_ptr<IGPUShader>, 4u> mainPipelineShaders = {};
-		constexpr auto vertexShaderPath = "../shaders/main_pipeline/vertex_shader.hlsl";
-		constexpr auto fragmentShaderPath = "../shaders/main_pipeline/fragment_shader.hlsl";
-		constexpr auto debugfragmentShaderPath = "../shaders/main_pipeline/fragment_shader_debug.hlsl";
-		constexpr auto resolveAlphasShaderPath = "../shaders/main_pipeline/resolve_alphas.hlsl";
-		// GeoTexture Pipeline Shaders
+
+		smart_refctd_ptr<IGPUShader> mainPipelineFragmentShaders = {};
+		smart_refctd_ptr<IGPUShader> mainPipelineVertexShader = {};
```
alichraghi, line 865: We had to copy if we wanted to compile vertex shader along with fragment shaders since we call `setShaderStage` to make pipeline creation happy. the proper fix would be letting `createXPipeline` take an optional shader stage field that overrides the `IShader` stage

devshgraphicsprogramming: ICPUShader has a `clone` for a reason, `IGPU` shall never be mutable.

devshgraphicsprogramming: can you do this stuff via clone as we discuessed, then I can merge the library stage stuff to `master`

devshgraphicsprogramming:
looks good, will merge right after  https://github.com/Devsh-Graphics-Programming/Nabla/pull/754

PR #181, opened 2025-01-29
Resizable cache

Fletterio:
Modifies ex 21 to use new resizable cache
https://github.com/Devsh-Graphics-Programming/Nabla/pull/823


21_LRUCacheUnitTest/main.cpp
```diff
@@ -77,16 +83,22 @@ class LRUCacheTestApp final : public nbl::application_templates::MonoSystemMonoL
 
 			cache.print(m_logger);
 
+			// Previous gets left the cahce with LRU order 13 11 10 12
```
Erfan-Ahmadi, line 86: a few key point when writing tests:
1. try two write small standalone blocks of code not depending on previous states, if it does, assert for it.
2. `assert` and not log, we need to `return false` + assert to programmaticly hint at a failure in our unit tests
  - logs are good but they shouldn't be the only tool + human understanding to validate the test. 
  - so your comments are good and may make someone understand this tests better, but add asserts 
 
taking care of the points above will make sure any future additions/edits by someone else on this file will not invalidate the tests or require them execute the code in their head for your tests to understand what should've happened.

IDEA: write one-liner state checking functions on the LRU cache. (to avoid cluttering the code after each `insert`)
something like:
```
assert(CheckCacheState(cache, {1, "2"}, {13, "L"}, ...));
```

> // Previous gets left the cahce with LRU order 13 11 10 12

here meaning: use a new cache, or assert that the previous state of the cache is at that state

21_LRUCacheUnitTest/main.cpp
```diff
@@ -145,6 +167,21 @@ class LRUCacheTestApp final : public nbl::application_templates::MonoSystemMonoL
 		#ifdef _NBL_DEBUG
 			cache2.print(m_logger);
 		#endif
+
+			// Grow test - try growing the cache
+			m_logger->log("Growing test");
```
Erfan-Ahmadi, line 172: call grow more for testing, and sometimes with a number lower than the previous.
(I think `grow` function returns bool, so check for it)

I know it's easy and you handled it, but you're not writing test for the thing you just wrote, you're writing tests for the junior programmer in 2027 trying to touch and modify the LRUCache.
i.e. make it impossible for anyone to pass this test after fucking with the LRUCache.

for example here if I fuck with you and make the grow funciton always fail, we will see no sign of failure, only different log.

TL;DR I'm nitpicking here because this is an important container, I wouldn't have been an*l about this if it was some other example or even test.

PR #186, opened 2025-03-25
CAD DTM Rendering

Erfan-Ahmadi:
CAD DTM Rendering Example
https://github.com/Devsh-Graphics-Programming/Nabla/pull/858


62_CAD/shaders/globals.hlsl
```diff
@@ -120,7 +126,7 @@ enum class MajorAxis : uint32_t
 struct MainObject
 {
     uint32_t styleIdx;
-    uint32_t pad; // do I even need this on the gpu side? it's stored in structured buffer not bda
+    uint32_t dtmSettingsIdx; // do I even need this on the gpu side? it's stored in structured buffer not bda
```
Erfan-Ahmadi, line 129: remove my comment, realize it was dumb, of course the pad was needed.

62_CAD/shaders/main_pipeline/common.hlsl
```diff
@@ -208,19 +216,28 @@ struct PSInput
...
+    float3 getScreenSpaceVertexPos(uint32_t vertexIndex) { return vertexScreenSpacePos[vertexIndex]; }
+#endif 
 };
 
 // Set 0 - Scene Data and Globals, buffer bindings don't change the buffers only get updated
 [[vk::binding(0, 0)]] ConstantBuffer<Globals> globals : register(b0);
 [[vk::binding(1, 0)]] StructuredBuffer<DrawObject> drawObjects : register(t0);
 [[vk::binding(2, 0)]] StructuredBuffer<MainObject> mainObjects : register(t1);
 [[vk::binding(3, 0)]] StructuredBuffer<LineStyle> lineStyles : register(t2);
+[[vk::binding(4, 0)]] StructuredBuffer<DTMSettings> dtmSettingsBuff : register(t3);
```
Erfan-Ahmadi, line 234: small rename, just `dtmSettings`

62_CAD/shaders/globals.hlsl
```diff
@@ -265,6 +271,12 @@ NBL_CONSTEXPR float InvalidStyleStretchValue = nbl::hlsl::numeric_limits<float>:
 // TODO[Przemek]: we will need something similar to LineStyles but related to heigh shading settings which is user customizable (like LineStyle stipple patterns) and requires upper_bound to figure out the color based on height value.
 // We'll discuss that later or what it will be looking like and how it's gonna get passed to our shaders.
 
+struct TriangleMeshVertex
```
Erfan-Ahmadi, line 274: If this is your vertex data you pull in vertex buffer,  it needs to be in double precision.
+ no need for this struct, just use a `pfloat64_t3` the z value is height

62_CAD/shaders/main_pipeline/vertex_shader.hlsl
```diff
@@ -92,6 +94,51 @@ PSInput main(uint vertexID : SV_VertexID)
...
+    outV.position.zw = float2(0.0, 1.0);
+    outV.data1 = uint4(0, 0, 0, 0);
+    outV.data2 = float4(0, 0, 0, 0);
+    outV.data3 = float4(0, 0, 0, 0);
+    outV.data4 = float4(0, 0, 0, 0);
+    outV.interp_data5 = float2(0, 0);
+    outV.setObjType(ObjectType::TRIANGLE_MESH);
+    outV.setMainObjectIdx(pc.triangleMeshMainObjectIndex);
+
+    TriangleMeshVertex vtx = vk::RawBufferLoad<TriangleMeshVertex>(pc.triangleMeshVerticesBaseAddress + sizeof(TriangleMeshVertex) * vertexID, 4u);
```
Erfan-Ahmadi, line 111: pull a `pfloat64_t3` and change alignment to `8u`

62_CAD/shaders/main_pipeline/vertex_shader.hlsl
```diff
@@ -92,6 +94,51 @@ PSInput main(uint vertexID : SV_VertexID)
...
+    vtxPos.y = _static_cast<pfloat64_t>(vtx.pos.y);
+
+    MainObject mainObj = mainObjects[pc.triangleMeshMainObjectIndex];
+    ClipProjectionData clipProjectionData = getClipProjectionData(mainObj);
+
+    float2 transformedPos = transformPointScreenSpace(clipProjectionData.projectionToNDC, globals.resolution, vtxPos);
+
+    outV.position.xy = transformedPos;
+    outV.position = transformFromSreenSpaceToNdc(outV.position.xy, globals.resolution);
+    outV.setHeightAtMeshVertex(vtx.height);
```
Erfan-Ahmadi, line 123: I need to think about this a little more. my intuition says height being in `float` is just fine.
So you could just cast the height value to float here, no need to worry about dealing with `.z` being double.
in the gpu-driven thing I'm doing I might make this height atrribute of each vertex a contigous separate block in floats instead of doubles.

for now just use `pfloat64_t3` for your vertices and cast the height to float.

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -408,11 +410,76 @@ float4 fragMain(PSInput input) : SV_TARGET
...
+        float3 v0 = input.getScreenSpaceVertexPos(0);
+        float3 v1 = input.getScreenSpaceVertexPos(1);
+        float3 v2 = input.getScreenSpaceVertexPos(2);
+
+        float2 start;
+        float2 end;
+        const float3 baryCoord = nbl::hlsl::spirv::BaryCoordKHR;
+
+        // TODO: figure out if branching can be reduced
+        // finding line start and end points by excluding vertex with the lowest barycentric coordinate value
```
Erfan-Ahmadi, line 428: you'll need to do SDF with potentially all 3 edges for the outline:
I've already covered this in our chat [here](https://discord.com/channels/593902898015109131/1338454683869515847/1351406137827921961)

```cpp
float triangleSDF = MAX_DISTANCE;
for (int n = 0; n < 3; ++n)
{
  float distanceToEdge = screenSpaceHeight[n]*baryCoord[n];
  float edgeSDF = abs(distanceToEdge) - thickness; 
  const bool edgeCanAffectCurrentPixel = edgeSDF <= aaFactor; 

  if (complexStyle && edgeCanAffectCurrentPixel)
  {
    edgeSDF = ...; // do complex line style with the custom lineStyleClipper and shit, you know the else case thing
  }

  triangleSDF = min(triangleSDF, edgeSDF);
}
```

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -408,11 +410,76 @@ float4 fragMain(PSInput input) : SV_TARGET
...
+        else if (baryCoord.z < baryCoord.x && baryCoord.z < baryCoord.y)
+        {
+            start = float2(v0.x, v0.y);
+            end = float2(v1.x, v1.y);
+        }
+
+        // long story short, in order for stipple patterns to be consistent:
+        // - point with lesser x coord should be starting point
+        // - if x coord of both points are equal then point with lesser y value should be starting point
+        if (end.x < start.x)
```
Erfan-Ahmadi, line 448: do this checks to see whether to use `end` or `start` when constructing the line (before doing sdf of it in https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/186/files#r2013523975)

62_CAD/CTriangleMesh.h
```diff
@@ -0,0 +1,66 @@
...
+
+	inline size_t getVtxBuffByteSize() const
+	{
+		return sizeof(vertex_t) * m_vertices.size();
+	}
+	inline size_t getIdxBuffByteSize() const
+	{
+		return sizeof(index_t) * m_indices.size();
+	}
+	inline size_t getIdxCnt() const
```
Erfan-Ahmadi, line 57: Vtx -> Vertex
Idx -> Index
Buff -> Buffer
Cnt -> Count
Please :)

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -84,7 +84,7 @@ void DrawResourcesFiller::allocateGeometryBuffer(ILogicalDevice* logicalDevice,
 
 	IGPUBuffer::SCreationParams geometryCreationParams = {};
 	geometryCreationParams.size = size;
-	geometryCreationParams.usage = bitflag(IGPUBuffer::EUF_STORAGE_BUFFER_BIT) | IGPUBuffer::EUF_SHADER_DEVICE_ADDRESS_BIT | IGPUBuffer::EUF_TRANSFER_DST_BIT;
+	geometryCreationParams.usage = bitflag(IGPUBuffer::EUF_STORAGE_BUFFER_BIT) | IGPUBuffer::EUF_SHADER_DEVICE_ADDRESS_BIT | IGPUBuffer::EUF_TRANSFER_DST_BIT | IGPUBuffer::EUF_INDEX_BUFFER_BIT;
```
Erfan-Ahmadi, line 87: add comment;
// INDEX_BUFFER USAGE for DTMs

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -218,6 +236,52 @@ void DrawResourcesFiller::drawPolyline(const CPolylineBase& polyline, uint32_t p
...
+	
+	// concatenate the index and vertex buffer into the geometry buffer
+	const size_t indexBuffByteSize = mesh.getIdxBuffByteSize();
+	const size_t vtxBuffByteSize = mesh.getVtxBuffByteSize();
+	const size_t geometryBufferDataToAddByteSize = indexBuffByteSize + vtxBuffByteSize;
+
+	// copy into gemoetry cpu buffer insteaed
+
+	// TODO: rename, its not just points
+	const uint32_t maxGeometryBufferPoints = static_cast<uint32_t>(maxGeometryBufferSize - currentGeometryBufferSize);
```
Erfan-Ahmadi, line 251: Rename to remainingGeometryBufferSize

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -218,6 +236,52 @@ void DrawResourcesFiller::drawPolyline(const CPolylineBase& polyline, uint32_t p
...
+
+	// TODO: assert of geometry buffer size, do i need to check if size of objects to be added <= maxGeometryBufferPoints?
+	// TODO: auto submit instead of assert
```
Erfan-Ahmadi, line 254: Good Todo comment, we'll tackle that later together

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -335,10 +400,27 @@ uint32_t DrawResourcesFiller::addLineStyle_SubmitIfNeeded(const LineStyleInfo& l
...
+uint32_t DrawResourcesFiller::addDTMSettings_SubmitIfNeeded(const DTMSettingsInfo& dtmSettings, SIntendedSubmitInfo& intendedNextSubmit)
+{
+	uint32_t outDTMSettingIdx = addDTMSettings_Internal(dtmSettings, intendedNextSubmit);
+	if (outDTMSettingIdx == InvalidStyleIdx)
+	{
+		finalizeAllCopiesToGPU(intendedNextSubmit);
+		submitDraws(intendedNextSubmit);
+		resetGeometryCounters();
+		resetMainObjectCounters();
+		resetLineStyleCounters();
```
Erfan-Ahmadi, line 412: I think here you also need a reset dtm settings counter

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -452,6 +535,26 @@ bool DrawResourcesFiller::finalizeLineStyleCopiesToGPU(SIntendedSubmitInfo& inte
 	return success;
 }
 
+bool DrawResourcesFiller::finalizeDTMSettingsCopiesToGPU(SIntendedSubmitInfo& intendedNextSubmit)
+{
+	bool success = true;
+	// Copy LineStyles
+	uint32_t remainingLineStyles = currentDTMSettingsCount - inMemDTMSettingsCount;
```
Erfan-Ahmadi, line 542: Comment and var rename needed here

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -684,6 +787,33 @@ uint32_t DrawResourcesFiller::addLineStyle_Internal(const LineStyleInfo& lineSty
 	return currentLineStylesCount++;
 }
 
+uint32_t DrawResourcesFiller::addDTMSettings_Internal(const DTMSettingsInfo& dtmSettingsInfo, SIntendedSubmitInfo& intendedNextSubmit)
+{
+	DTMSettings dtmSettings;
+
+	// TODO: this needs to be redone.. what if submit happens after that line?
+	// we need to make sure somehow that function below will not submit, we need both outline and contour styles in GPU memory
+	dtmSettings.outlineLineStyleIdx = addLineStyle_SubmitIfNeeded(dtmSettingsInfo.outlineLineStyleInfo, intendedNextSubmit);
```
Erfan-Ahmadi, line 796: Asset here that the max line styles is >= 2, because you need at least 2 resident lineStyles. And you don't want the second call to evict the first

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -684,6 +787,33 @@ uint32_t DrawResourcesFiller::addLineStyle_Internal(const LineStyleInfo& lineSty
 	return currentLineStylesCount++;
 }
 
+uint32_t DrawResourcesFiller::addDTMSettings_Internal(const DTMSettingsInfo& dtmSettingsInfo, SIntendedSubmitInfo& intendedNextSubmit)
+{
+	DTMSettings dtmSettings;
+
+	// TODO: this needs to be redone.. what if submit happens after that line?
+	// we need to make sure somehow that function below will not submit, we need both outline and contour styles in GPU memory
```
Erfan-Ahmadi, line 795: Yes you're right, even if you have many line styles and the first one fits into men, the second one might evict everything. 
In the case that remainingLineStyles is < 2, do the auto submit calls and resets. And then call addLineStyle_Internal for both (asset they succeed, because they definitely should)

Also be careful to not call this in the middle of adding geometry that should reference this dtm settings.

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -684,6 +787,33 @@ uint32_t DrawResourcesFiller::addLineStyle_Internal(const LineStyleInfo& lineSty
...
+
+	DTMSettings* settingsArray = reinterpret_cast<DTMSettings*>(cpuDrawBuffers.dtmSettingsBuffer->getPointer());
+	for (uint32_t i = 0u; i < currentDTMSettingsCount; ++i)
+	{
+		const DTMSettings& itr = settingsArray[i];
+		if (itr == dtmSettings)
+			return i;
+	}
+
+	if (currentDTMSettingsCount >= maxDtmSettings)
```
Erfan-Ahmadi, line 807: We should be extra careful and document more robustly. 
Auto submit can easily break with multiple _SubmitIfNeeded calls.
I'm sure I've stated explicitly somewhere in the comments that user shall not batch calls to  addMainObj.
Make sure you add the same comments to line styles and dtm settings (is shouldn't be called twice or more in a row and then used via index)

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -218,6 +236,52 @@ void DrawResourcesFiller::drawPolyline(const CPolylineBase& polyline, uint32_t p
...
+	// copy into gemoetry cpu buffer insteaed
+
+	// TODO: rename, its not just points
+	const uint32_t maxGeometryBufferPoints = static_cast<uint32_t>(maxGeometryBufferSize - currentGeometryBufferSize);
+
+	// TODO: assert of geometry buffer size, do i need to check if size of objects to be added <= maxGeometryBufferPoints?
+	// TODO: auto submit instead of assert
+	assert(geometryBufferDataToAddByteSize <= maxGeometryBufferPoints);
+
+	// TODO: vertices need to be aligned to 8?
```
Erfan-Ahmadi, line 257: Yes I probably missed this in other places, we need to align to 8 bytes because the vertices have doubles in them. Keep this comment, I'll come back to it

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -218,6 +236,52 @@ void DrawResourcesFiller::drawPolyline(const CPolylineBase& polyline, uint32_t p
...
+	uint64_t vtxBufferAddress;
+	{
+		void* dst = reinterpret_cast<char*>(cpuDrawBuffers.geometryBuffer->getPointer()) + currentGeometryBufferSize;
```
Erfan-Ahmadi, line 260: uint8_t please no char. And why are you reinterpret casting and storing as void*

Erfan-Ahmadi: Just store as uin8_t*, and advance it with+= and use the same pointer for the next copy

62_CAD/shaders/globals.hlsl
Przemog1: does it matter when its this struct or `pfloat64_t3`?

Przemog1: if performance wise it doesn't matter then i would prefer to keep the struct

Przemog1: it makes code a little bit more self explanatory

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
Przemog1: outdated, resolving

62_CAD/CTriangleMesh.h
```diff
@@ -0,0 +1,131 @@
...
+	LineStyleInfo lineStyleInfo;
+
+	float startHeight;
+	float endHeight;
+	float heightInterval;
+};
+
+struct DTMSettingsInfo
+{
+	uint32_t mode = 0u;
```
Erfan-Ahmadi, line 79: add comment that it related to E_DTM_MODE

62_CAD/shaders/globals.hlsl
```diff
@@ -262,9 +294,15 @@ NBL_CONSTEXPR uint32_t InvalidRigidSegmentIndex = 0xffffffff;
...
 
 
-// TODO[Przemek]: we will need something similar to LineStyles but related to heigh shading settings which is user customizable (like LineStyle stipple patterns) and requires upper_bound to figure out the color based on height value.
+// TODO[Przemek]: we will need something similar to LineStyles but related to heigh shading settings which is user customizable (like  stipple patterns) and requires upper_bound to figure out the color based on height value.
 // We'll discuss that later or what it will be looking like and how it's gonna get passed to our shaders.
 
+struct TriangleMeshVertex
+{
+    pfloat64_t2 pos;
+    pfloat64_t height;
```
Erfan-Ahmadi, line 303: add comment here that `height` can be `float32_t`.
I'll later handle in my gpu driven work, right now it wouldn't help with size of the struct due to padding, unless stored in a separate array (SoA style)

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -1,655 +1,1104 @@
...
+#include "common.hlsl"
+#include <nbl/builtin/hlsl/shapes/beziers.hlsl>
+#include <nbl/builtin/hlsl/shapes/line.hlsl>
+#include <nbl/builtin/hlsl/algorithm.hlsl>
+#include <nbl/builtin/hlsl/math/equations/quadratic.hlsl>
+#include <nbl/builtin/hlsl/math/geometry.hlsl>
+#include <nbl/builtin/hlsl/spirv_intrinsics/fragment_shader_pixel_interlock.hlsl>
+#include <nbl/builtin/hlsl/jit/device_capabilities.hlsl>
+#include <nbl/builtin/hlsl/text_rendering/msdf.hlsl>
+#include <nbl/builtin/hlsl/spirv_intrinsics/fragment_shader_barycentric.hlsl>
```
Erfan-Ahmadi, line 11: do you use this anymoe? maybe comment it out for now and say we compute the barycentric ourselves due to dilation

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -1,655 +1,1104 @@
...
+    return color;
+}
+
+float dot2(in float2 vec)
+{
+    return dot(vec, vec);
+}
+
+
+// TODO: Later move these functions and structs to dtmSettings.hlsl and a namespace like dtmSettings::height_shading or dtmSettings::contours, etc..
```
Erfan-Ahmadi, line 424: Please move all the dtm related, fragment shader functions, including the SDFs to another file called `dtm.hlsl` in `namespace dtm`
I don't like how cluttered the fragment_shader is becoming.

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -1,655 +1,1104 @@
...
+
+    }
+
+    outputColor.a = smoothstep(+globals.antiAliasingFactor, -globals.antiAliasingFactor, minDistance) * outlineStyle.color.a;
+    outputColor.rgb = outlineStyle.color.rgb;
+
+    return outputColor;
+}
+
+float4 blendColorOnTop(in float4 colorBelow, in float4 colorAbove)
```
Erfan-Ahmadi, line 810: use blending jargon, srcColor, dstColor

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -1,655 +1,1104 @@
...
+{
+    float4 outputColor = colorBelow;
+    outputColor.rgb = colorAbove.rgb * colorAbove.a + outputColor.rgb * outputColor.a * (1.0f - colorAbove.a);
+    outputColor.a = colorAbove.a + outputColor.a * (1.0f - colorAbove.a);
```
Erfan-Ahmadi, line 814: There are a few notes regarding this custom blending which I believe might be the cause of some bugs:

- first ever blendColorOnTop, should override outputColor (dtmColor) and not blend at all. you're wrongly blending with color black (let's say your only have height shading and it's color is (1, 0, 0, 0.5)), background color can be anything. you should let the hardware do the final blend.

- We either needs to max the alphas or replace with srcAlpha(colorAbove.a) completely. I need to think about what makes more sense.

- It's correct that we're using srcAlpha to blend dst and src color, but it only makes sense if we blend that after it's blended with the background.

TL;DR I need some time to figure this shit out

Erfan-Ahmadi: Ok we should do under blending, change this function's name to `blendUnder` and then render the other way around.
and as an optimization skip rendering the next stuff if alpha is 1.0 (the color is gonna get discarded for things behind)

the under blending operator is as follows:
```
dst.rgb = dst.rgb*dst.a + (1-dst.a)*src.a*src.rgb; // src and dst swapped
dst.a = (1.0f-src.a)*dst.a + src.a; // like currently implemented
``
the current implementation while it looks correct, fails when you just blend a bunch of fully transparent things on top of a semi transparent thing because `outputColor.rgb = outputColor.rgb*outputColor.a;` each time.

File: 62_CAD/shaders/main_pipeline/fragment_shader.hlsl [hunk 25]
```diff
@@ -1,655 +1,1100 @@
...
+
+        return outputColor;
+    }
+    
+    return float4(0.0f, 0.0f, 0.0f, 0.0f);
+}
+
+float4 calculateDTMOutlineColor(in DTMSettings dtmSettings, in float3 v[3], in uint2 edgePoints[3], in PSInput psInput, in float3 baryCoord, in float height)
+{
+    float4 outputColor;
```
Erfan-Ahmadi, line 716: default initialize

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -1,655 +1,1100 @@
...
+    {
+        const uint2 currentEdgePoints = edgePoints[i];
+        float3 A = v[currentEdgePoints[0]];
+        float3 B = v[currentEdgePoints[1]];
+        float3 AB = B - A;
+        float ABLen = length(AB);
+        float triangleHeightToOpositeVertex = triangleAreaTimesTwo / ABLen;
+
+        distances[i] = abs(triangleHeightToOpositeVertex * baryCoord[opposingVertexIdx[i]]);
+    }
```
Erfan-Ahmadi, line 754: seems like you don't need these anymore

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -1,655 +1,1100 @@
...
+        DTMSettings dtmSettings = loadDTMSettings(mainObj.dtmSettingsIdx);
+
+        float3 v[3];
+        v[0] = input.getScreenSpaceVertexAttribs(0);
+        v[1] = input.getScreenSpaceVertexAttribs(1);
+        v[2] = input.getScreenSpaceVertexAttribs(2);
+
+        // indices of points constructing every edge
+        uint2 edgePoints[3];
+        edgePoints[0] = uint2(0, 1);
```
Erfan-Ahmadi, line 837: why not just use `i` and `(i+1)%3`
will make the code clean and more readable

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -1,655 +1,1104 @@
...
+    float stretch = 1.0f;
+    float phaseShift = 0.0f;
+
+    // TODO: move to ubo or push constants
+    const float startHeight = contourSettings.contourLinesStartHeight;
+    const float endHeight = contourSettings.contourLinesEndHeight;
+    const float interval = contourSettings.contourLinesHeightInterval;
+
+    // TODO: can be precomputed
+    const int maxContourLineIdx = (endHeight - startHeight + 1) / interval;
```
Erfan-Ahmadi, line 653: What's the +1 here? You're working with floating point values. 
Also it's wrong when start=0, end=10 and interval=5.5, it will give wrong maxIdx
I think It should work without it?

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -1,655 +1,1104 @@
...
+
+    // TODO: it actually can output a negative number, fix
+    int contourLineIdx = nbl::hlsl::_static_cast<int>((height - startHeight + (interval * 0.5f)) / interval);
```
Erfan-Ahmadi, line 656: (interval*0.5f)/interval is just 0.5f, 
you can just do `(height - startHeight) / interval + 0.5f`

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -1,655 +1,1104 @@
...
+    for (int i = 0; i < 3; ++i)
+    {
+        if (contourLinePointsIdx == 2)
+            break;
+
+        const uint2 currentEdgePoints = edgePoints[i];
+        float3 p0 = v[currentEdgePoints[0]];
+        float3 p1 = v[currentEdgePoints[1]];
+
+        if (p1.z < p0.z)
```
Erfan-Ahmadi, line 672: comment on why you're doing this (stippling)

62_CAD/shaders/main_pipeline/dtm.hlsl
```diff
@@ -0,0 +1,678 @@
+#ifndef _CAD_EXAMPLE_DTM_HLSL_INCLUDED_
+#define _CAD_EXAMPLE_DTM_HLSL_INCLUDED_
+
+#include <nbl/builtin/hlsl/shapes/line.hlsl>
+#include <nbl/builtin/hlsl/algorithm.hlsl>
+
+// TODO: functions outside of the "dtm" namespace need to be moved to another file
+
+// for usage in upper_bound function
+struct StyleAccessor
```
Erfan-Ahmadi, line 10: move these to line_style.hlsl, and include it here
these aren't related to dtms

62_CAD/shaders/main_pipeline/dtm.hlsl
```diff
@@ -0,0 +1,678 @@
...
+        return settings.heightColorMapHeights[ix];
+    }
+};
+
+float dot2(in float2 vec)
+{
+    return dot(vec, vec);
+}
+
+// TODO: Later move these functions and structs to dtmSettings.hlsl and a namespace like dtmSettings::height_shading or dtmSettings::contours, etc..
```
Erfan-Ahmadi, line 319: remove comment

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
Erfan-Ahmadi: Edits, after putting some thought into it and reading [dual depth](https://developer.download.nvidia.com/SDK/10/opengl/src/dual_depth_peeling/doc/DualDepthPeeling.pdf), we should replace `dst.rgb*dst.a` with just `dst.rgb` because the multiplication was taken into account in previous steps by `src.a*src.rgb`

so the whole expression becomes:
```
dst.rgb = dst.rgb + (1-dst.a)*src.a*src.rgb; // src and dst swapped
dst.a = (1.0f-src.a)*dst.a + src.a; // like currently implemented
```

test if this fixes your issue [here](https://discord.com/channels/593902898015109131/1338454683869515847/1364345059109240943)
don't forget to reverse the drawing order.

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -309,256 +360,683 @@ void DrawResourcesFiller::drawFontGlyph(
...
 	}
-	return success;
+
+	setActiveDTMSettings(dtmSettingsInfo);
+	beginMainObject(MainObjectType::GRID_DTM);
+
+	uint32_t mainObjectIdx = acquireActiveMainObjectIndex_SubmitIfNeeded(intendedNextSubmit);
+	assert(mainObjectIdx != InvalidMainObjectIdx);
+
+	addGridDTM_Internal(gridDTMInfo, mainObjectIdx);
```
Erfan-Ahmadi, line 613: you need to follow this pattern to handle possible overflow submit (in case there is not enough mem for GridDTMInfo)
```c++
	if (!addImageObject_Internal(info, mainObjIdx))
	{
		// single image object couldn't fit into memory to push to gpu, so we submit rendering current objects and reset geometry buffer and draw objects
		submitCurrentDrawObjectsAndReset(intendedNextSubmit, mainObjIdx);
		bool success = addImageObject_Internal(info, mainObjIdx);
		assert(success); // this should always be true, otherwise it's either bug in code or not enough memory allocated to hold a single image object 
	}
```

62_CAD/shaders/main_pipeline/vertex_shader.hlsl
```diff
@@ -107,486 +118,600 @@ PSInput main(uint vertexID : SV_VertexID)
...
+            pfloat64_t height = vk::RawBufferLoad<pfloat64_t>(globals.pointers.geometryBuffer + drawObj.geometryAddress + sizeof(pfloat64_t2), 8u);
+            pfloat64_t width = vk::RawBufferLoad<pfloat64_t>(globals.pointers.geometryBuffer + drawObj.geometryAddress + sizeof(pfloat64_t2) + sizeof(pfloat64_t), 8u);
+            uint32_t dtmSettingsID = vk::RawBufferLoad<uint32_t>(globals.pointers.geometryBuffer + drawObj.geometryAddress + sizeof(pfloat64_t2) + 2 * sizeof(pfloat64_t), 8u);
+            float gridCellWidth = vk::RawBufferLoad<float>(globals.pointers.geometryBuffer + drawObj.geometryAddress + sizeof(pfloat64_t2) + 2 * sizeof(pfloat64_t) + sizeof(uint32_t), 8u);
+            float reciprocalOutlineStipplePatternLength = vk::RawBufferLoad<float>(globals.pointers.geometryBuffer + drawObj.geometryAddress + sizeof(pfloat64_t2) + 2 * sizeof(pfloat64_t) + sizeof(uint32_t) + sizeof(float), 8u);
+
+            const float2 corner = float2(bool2(vertexIdx & 0x1u, vertexIdx >> 1));
+            pfloat64_t2 vtxPos = topLeft;
+            if (corner.x)
+                vtxPos.x = vtxPos.x + width;
```
Erfan-Ahmadi, line 660: I would do `vtxPos = vtxPos + corner * (width, height);`

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -1,655 +1,602 @@
...
+                float2 gridSpacePos = uv * gridExtents;
+
+                float2 cellCoords;
+                {
+                    float2 gridSpacePosDivGridCellWidth = gridSpacePos / cellWidth;
+                    cellCoords.x = uint32_t(gridSpacePosDivGridCellWidth.x);
+                    cellCoords.y = uint32_t(gridSpacePosDivGridCellWidth.y);
+                }
+
+                float2 insideCellCoord = gridSpacePos - float2(cellWidth, cellWidth) * cellCoords; // TODO: use fmod instead?
```
Erfan-Ahmadi, line 475: @devshgraphicsprogramming what do you think about adding `OpFMod` to our builtins?

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -1,655 +1,602 @@
...
+                
+                const float InvalidHeightValue = asfloat(0x7FC00000);
```
Erfan-Ahmadi, line 477: make this a global variable in globals.hlsl with an accompanying `isInvalidHeightValue` we might do a comparison check  against a very low value instead of using NaN

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -1,655 +1,602 @@
...
+}
+
+enum E_CELL_DIAGONAL
+{
+    TOP_LEFT_TO_BOTTOM_RIGHT,
+    BOTTOM_LEFT_TO_TOP_RIGHT,
+    INVALID
+};
+
+E_CELL_DIAGONAL resolveGridDTMCellDiagonal(in float4 cellHeights)
```
Erfan-Ahmadi, line 127: move these to dtm header

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -1,655 +1,602 @@
...
+{
+    static const E_CELL_DIAGONAL DefaultDiagonal = TOP_LEFT_TO_BOTTOM_RIGHT;
+
+    const bool4 invalidHeights = bool4(
+        isnan(cellHeights.x),
```
Erfan-Ahmadi, line 132: use IsInvalid from the new globals.hlsl function I mentioned in https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/186/files#r2133674200

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -1,655 +1,602 @@
...
+
+                    cellHeights = textures[NonUniformResourceIndex(textureId)].Gather(textureSampler, float2(location.x, location.y), 0);
+                }
+
+
+                const E_CELL_DIAGONAL cellDiagonal = resolveGridDTMCellDiagonal(cellHeights);
+                const bool diagonalFromTopLeftToBottomRight = cellDiagonal == E_CELL_DIAGONAL::TOP_LEFT_TO_BOTTOM_RIGHT;
+
+                if (cellDiagonal == E_CELL_DIAGONAL::INVALID)
+                    discard;
```
Erfan-Ahmadi, line 492: I think you're going to handle it when doing the dilation emulation (when you iterate over all neighbouring cells)
but I'm going to say it again anyways (sorry): we shouldn't discard

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -1,655 +1,602 @@
...
+                {
+                    float2 gridSpacePosDivGridCellWidth = gridSpacePos / cellWidth;
+                    cellCoords.x = uint32_t(gridSpacePosDivGridCellWidth.x);
+                    cellCoords.y = uint32_t(gridSpacePosDivGridCellWidth.y);
+                }
+
+                float2 insideCellCoord = gridSpacePos - float2(cellWidth, cellWidth) * cellCoords; // TODO: use fmod instead?
+                
+                const float InvalidHeightValue = asfloat(0x7FC00000);
+                float4 cellHeights = float4(InvalidHeightValue, InvalidHeightValue, InvalidHeightValue, InvalidHeightValue);
```
Erfan-Ahmadi, line 478: Comment the formation: is it (TL, TR, BL, BR) for example?

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -1,655 +1,602 @@
...
+                if (triangleA)
+                {
+                    outlineLineSegments[0] = nbl::hlsl::shapes::Line<float>::construct(v[2].xy, v[0].xy);
+                    outlineLineSegments[1] = nbl::hlsl::shapes::Line<float>::construct(v[2].xy, v[1].xy);
+                }
+                else
+                {
+                    outlineLineSegments[0] = nbl::hlsl::shapes::Line<float>::construct(v[1].xy, v[2].xy);
+                    outlineLineSegments[1] = nbl::hlsl::shapes::Line<float>::construct(v[0].xy, v[2].xy);
+                }
```
Erfan-Ahmadi, line 536: Ok so this is not the way I want you to do outlines at all.
1. Triangle formation (your v[3]) should only be done for CONTOUR and HEIGHT_SHADING (skip tthe formation calculations if only outlines is requested)
2. Outline SDF and color calculation should be independant from triangle formation, based on the point you're going to do SDF with (2 lines out of 4)
3. You'll be doing sdf with the whole larger lines, you don't have to adjust phaseshift anymore (extra unnecessary calculations), we'll handle clipping for all types of objects later on in case it was super zoomed in the sdf might be less accurate, but it's a simple line sdf so no big problem.
4. Remember that the whole dilation emulation thing (iterating over all neighbouring cells and doing sdf with them) should also be independant from outline rendering
5. Last but not least I want to be able to draw a grid without a heightmap texture and triangle formation. make sure your code can handle that. Because we're going to abuse this code to draw simple regular grids in n4ce, where there is no height values and heightmaps, formations or triangles; just simple grid
image

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -1,655 +1,602 @@
...
+            const bool outOfBoundsUV = uv.x < 0.0f || uv.y < 0.0f || uv.x > 1.0f || uv.y > 1.0f;
+            float4 dtmColor = float4(0.0f, 0.0f, 0.0f, 0.0f);
+            if (dtmSettings.drawContourEnabled())
+            {
+                for (int i = dtmSettings.contourSettingsCount-1u; i >= 0; --i) 
+                    dtmColor = dtm::blendUnder(dtmColor, dtm::calculateDTMContourColor(dtmSettings.contourSettings[i], v, input.position.xy, height));
+            }
+            if (dtmSettings.drawOutlineEnabled())
+                dtmColor = dtm::blendUnder(dtmColor, dtm::calculateGridDTMOutlineColor(dtmSettings.outlineLineStyleIdx, outlineLineSegments, input.position.xy, outlinePhaseShift));
+            if (dtmSettings.drawHeightShadingEnabled() && !outOfBoundsUV)
```
Erfan-Ahmadi, line 566: Ok I can live with this. just know that you're implicitly discarding height shading for out of bounds UVs. if your grid was rotated you'd get aliasing.

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -1,655 +1,602 @@
...
+            // v2a-------v1   v0-------v2b
+            // |  A     / |   | \     B  |
+            // |     /    |   |    \     |
+            // |  /  B    |   |   A   \  |
+            // v0-------v2b   v2a-------v1
+            // 
+
+            // calculate screen space coordinates of vertices of t
+            // he current tiranlge within the grid
+            float3 v[3];
```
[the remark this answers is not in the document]
Przemog1: for now i only changed name of the variable you highlighted, honestly i doubt ai can be clever enough to understand the code and give reasonable varable names but maybe i will try someday xd

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
Przemog1: moved `InvalidHeightValue` to globals, but i'm not sure what you mean by
> with an accompanying `isInvalidHeightValue`

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
Przemog1: sure

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
Przemog1: i don't get it

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
Przemog1: no, it needs to be fixed, it causes this:
image

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -780,154 +1957,528 @@ void DrawResourcesFiller::addLines_Internal(const CPolylineBase& polyline, const
...
+
+	// Attempt to create a GPU image and corresponding image view for this texture.
+	// If creation or memory allocation fails (likely due to VRAM exhaustion),
+	// we'll evict another texture from the LRU cache and retry until successful, or until only the currently-cachedImageRecord image remains.
+	while (imagesCache->size() > 0u)
+	{
+		// Try creating the image and allocating memory for it:
+		nbl::video::IGPUImage::SCreationParams params = {};
+		params = imageParams;
+		auto gpuImage = device->createImage(std::move(params));
```
Erfan-Ahmadi, line 2343: Where do you add the `ECF_MUTABLE_FORMAT_BIT` flag?  I think it should be added when override is valid and not equal to ICPUImage format (the image creation params given).
It shouldn't be user's job to add it

Erfan-Ahmadi: Add the view format override in `params.viewFormat` (if valid and inequal to image format).
because we want the image creation to fail if we cannot create view with that format for the image.

Przemog1: > Add the view format override in `params.viewFormat` (if valid and inequal to image format). because we want the image creation to fail if we cannot create view with that format for the image.

TODO:
https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/blob/dtm/62_CAD/DrawResourcesFiller.cpp#L2350-L2351

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -1,655 +1,602 @@
...
+                    v[0] = float3(gridSpaceCellTopLeftCoords.x, gridSpaceCellTopLeftCoords.y, cellHeights.w);
+                    v[1] = float3(gridSpaceCellTopLeftCoords.x + cellWidth, gridSpaceCellTopLeftCoords.y + cellWidth, cellHeights.y);
+                    v[2] = triangleA ? float3(gridSpaceCellTopLeftCoords.x, gridSpaceCellTopLeftCoords.y + cellWidth, cellHeights.x) : float3(gridSpaceCellTopLeftCoords.x + cellWidth, gridSpaceCellTopLeftCoords.y, cellHeights.z);
+                }
+                else
+                {
+                    v[0] = float3(gridSpaceCellTopLeftCoords.x, gridSpaceCellTopLeftCoords.y + cellWidth, cellHeights.x);
+                    v[1] = float3(gridSpaceCellTopLeftCoords.x + cellWidth, gridSpaceCellTopLeftCoords.y, cellHeights.z);
+                    v[2] = triangleA ? float3(gridSpaceCellTopLeftCoords.x, gridSpaceCellTopLeftCoords.y, cellHeights.w) : float3(gridSpaceCellTopLeftCoords.x + cellWidth, gridSpaceCellTopLeftCoords.y + cellWidth, cellHeights.y);
+                }
```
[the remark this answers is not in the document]
Przemog1: TODO

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -282,194 +453,758 @@ void DrawResourcesFiller::drawFontGlyph(
...
+// TODO[Przemek]: similar to other drawXXX and drawXXX_internal functions that create mainobjects, drawObjects and push additional info in geometry buffer, input to function would be a GridDTMInfo
+// We don't have an allocator or memory management for texture updates yet, see how `_test_addImageObject` is being temporarily used (Descriptor updates and pipeline barriers) to upload an image into gpu and update a descriptor slot (it will become more sophisticated but doesn't block you)
+void DrawResourcesFiller::drawGridDTM(
+	const float64_t2& topLeft,
+	float64_t2 worldSpaceExtents,
+	float gridCellWidth,
+	uint64_t textureID,
+	const DTMSettingsInfo& dtmSettingsInfo,
+	SIntendedSubmitInfo& intendedNextSubmit,
+	bool drawGridOnly/* = false*/)
```
Erfan-Ahmadi, line 780: remove this param and use dtmSettings mode instead. if it's OUTLINE then it's GRID only.
and texture ID isn't even considered. make sure that's documented in DrawResources.h.

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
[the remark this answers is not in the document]
Erfan-Ahmadi: agreed, we did this at a higher level. and no more top left addition needed

PR #195, opened 2025-06-14
Rework pipeline shader spec info

kevyuu:


67_RayQueryGeometry/app_resources/render.comp.hlsl
```diff
@@ -6,13 +6,25 @@
...
 
+
 using namespace nbl::hlsl;
 
 [[vk::push_constant]] SPushConstants pc;
 
 [[vk::binding(0, 0)]] RaytracingAccelerationStructure topLevelAS;
 
 [[vk::binding(1, 0)]] RWTexture2D<float4> outImage;
+[[vk::constant_id(0)]] const float shader_variant = 1.0;
```
devshgraphicsprogramming, line 17: why float and not `uint32_t` ?

PR #205, opened 2025-07-24
Implementing streamed tiles system for big georeferenced images

Fletterio:


62_CAD/DrawResourcesFiller.cpp
```diff
@@ -705,17 +704,17 @@ bool DrawResourcesFiller::ensureGeoreferencedImageAvailability_AllocateIfNeeded(
...
 			{
-				cachedImageRecord->type = georeferenceImageType;
+				cachedImageRecord->type = manager.imageType;
 				cachedImageRecord->state = ImageState::CREATED_AND_MEMORY_BOUND;
 				cachedImageRecord->lastUsedFrameIndex = currentFrameIndex; // there was an eviction + auto-submit, we need to update AGAIN
 				cachedImageRecord->allocationOffset = allocResults.allocationOffset;
 				cachedImageRecord->allocationSize = allocResults.allocationSize;
 				cachedImageRecord->gpuImageView = allocResults.gpuImageView;
-				cachedImageRecord->staticCPUImage = nullptr;
+				cachedImageRecord->staticCPUImage = manager.georeferencedImageParams.geoReferencedImage;
```
Erfan-Ahmadi, line 717: this should stay nullptr, georeferenced images should have nothing to do with staticCPUImage. the cacheImageRecord only holds the status of the gpu allocated georef image.
then you just copy your regions with buffers into it.

Erfan-Ahmadi: i.e.  the cpu image should only be part of the loader and not the draw resources filler OR the tile calculator

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -631,20 +631,19 @@ bool DrawResourcesFiller::ensureMultipleStaticImagesAvailability(std::span<Stati
 	return true;
 }
 
-bool DrawResourcesFiller::ensureGeoreferencedImageAvailability_AllocateIfNeeded(image_id imageID, const GeoreferencedImageParams& params, SIntendedSubmitInfo& intendedNextSubmit)
+bool DrawResourcesFiller::ensureGeoreferencedImageAvailability_AllocateIfNeeded(StreamedImageManager& manager, SIntendedSubmitInfo& intendedNextSubmit)
```
Erfan-Ahmadi, line 634: Ok so I have a few requests about this.
This function should only care about the GPU allocation and the image itself and decide which resolution to create it with. (+ if it's better to go full res or streamed)
And you can know these with only 1. OriginalImageResolution 2. ViewportResolution 3. TileSize 
I don't see any need for the manager to go through here.
If this function is doing more than just figuring out what resolution to allocate the image in the vram with, then it's a sign something is wrong.
-----
then when you're in `addGeorefImage` function, you can look up the cached image and see if it was decided to be "Stremead" or "FullRes" depending on that, you either request loading the whole shit from the loader OR you request tiles (through your tile manager or whatever)

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -2461,30 +2470,43 @@ DrawResourcesFiller::ImageAllocateResults DrawResourcesFiller::tryCreateAndAlloc
...
-	if (outImageType == ImageType::GEOREFERENCED_FULL_RESOLUTION)
+	if (manager.imageType == ImageType::GEOREFERENCED_FULL_RESOLUTION)
 	{
 		outImageParams.extent = { georeferencedImageParams.imageExtents.x, georeferencedImageParams.imageExtents.y, 1u };
 	}
 	else
 	{
-		// TODO: Better Logic, area around the view, etc...
-		outImageParams.extent = { georeferencedImageParams.viewportExtents.x, georeferencedImageParams.viewportExtents.y, 1u };
+		// Pad sides to multiple of tileSize. Even after rounding up, we might still need to add an extra tile to cover both sides.
```
Erfan-Ahmadi, line 2496: continuing from https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/205/files#r2242153374
I believe these calculations is better if done inside `addGeorefImage` function

62_CAD/main.cpp
```diff
@@ -361,6 +364,173 @@ bool performImageFormatPromotionCopy(const core::smart_refctd_ptr<asset::ICPUIma
...
+		return mipLevels[0]->getCreationParameters().format;
+	}
+
+private:
+	// These are here for the example, might not be class members when porting to n4ce
+	asset::IAssetManager* m_assetMgr = {};
+	system::ILogger* m_logger = {};
+	video::IPhysicalDevice* m_physicalDevice = {};
+	// We're going to fake it in the example so it's easier to work with, but the interface remains
+	core::vector<smart_refctd_ptr<ICPUImage>> mipLevels = {};
```
Erfan-Ahmadi, line 531: Unurgent + Unimportant (but I have to ask): 
an ICPUImage can have mip levels inside it. 
and why load images separately? I think `.exr` can hold multiple mip levels by itself and our loader can handle that.

62_CAD/main.cpp
```diff
@@ -494,10 +664,11 @@ class ComputerAidedDesign final : public nbl::examples::SimpleWindowedApplicatio
 		
 		// Static Image Sampler
 		{
+			constexpr auto wrapMode = mode == ExampleMode::CASE_12 ? IGPUSampler::E_TEXTURE_CLAMP::ETC_REPEAT : IGPUSampler::E_TEXTURE_CLAMP::ETC_MIRROR;
```
Erfan-Ahmadi, line 667: no need for this if it didn't mess up other examples

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -729,13 +729,23 @@ bool DrawResourcesFiller::ensureGeoreferencedImageAvailability_AllocateIfNeeded(
...
 				cachedImageRecord->state = ImageState::CREATED_AND_MEMORY_BOUND;
 				cachedImageRecord->lastUsedFrameIndex = currentFrameIndex; // there was an eviction + auto-submit, we need to update AGAIN
 				cachedImageRecord->allocationOffset = allocResults.allocationOffset;
 				cachedImageRecord->allocationSize = allocResults.allocationSize;
 				cachedImageRecord->gpuImageView = allocResults.gpuImageView;
 				cachedImageRecord->staticCPUImage = nullptr;
+				cachedImageRecord->georeferencedImageState = GeoreferencedImageStreamingState::create(std::move(params));
+
+				// This is because gpu image is square
+				cachedImageRecord->georeferencedImageState->maxResidentTiles = cachedImageRecord->gpuImageView->getCreationParameters().image->getCreationParameters().extent.width / GeoreferencedImageTileSize;
```
Erfan-Ahmadi, line 742: This is assuming width = height (due to rotation)?

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -729,13 +729,23 @@ bool DrawResourcesFiller::ensureGeoreferencedImageAvailability_AllocateIfNeeded(
...
+				
+				auto& maxImageTileIndices = cachedImageRecord->georeferencedImageState->maxImageTileIndices;
+				maxImageTileIndices = cachedImageRecord->georeferencedImageState->georeferencedImageParams.imageExtents / uint32_t2(GeoreferencedImageTileSize, GeoreferencedImageTileSize);
+				// If it fits perfectly along any dimension, we need one less tile with this scheme
+				maxImageTileIndices -= uint32_t2(maxImageTileIndices.x * GeoreferencedImageTileSize == cachedImageRecord->georeferencedImageState->georeferencedImageParams.imageExtents.x,
```
Erfan-Ahmadi, line 747: so maxResidentTiles is related to the gpu image we allocated it's only a single number because it's square.
maxResidentTiles is a bit misleading, maybe name that to contain "sideLength" or "dimension" in it's name.
I assume maxResidentTiles would be the number of resident tiles not sqrt of it

Erfan-Ahmadi, line 747: feels like you could just do `(imageExtents - 1) / TileSize` instead of this bool checker and -1, if it's exactly divisible by TileSize it gives you what you want.

it's very similar we decide how many groups to dispatch a compute shader with ( it's not Count/GroupSize but (Count-1)/GroupSize+1

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -897,11 +907,28 @@ void DrawResourcesFiller::addGeoreferencedImage(image_id imageID, const Georefer
 		return;
 	}
 
+	// Query imageType
+	auto cachedImageRecord = imagesCache->peek(imageID);
+	if (!cachedImageRecord)
+	{
+		m_logger.log("addGeoreferencedImage was not called immediately after enforceGeoreferencedImageAvailability!", nbl::system::ILogger::ELL_ERROR);
+		assert(false);
```
Erfan-Ahmadi, line 915: return also, there is no point in proceeding

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -897,11 +907,28 @@ void DrawResourcesFiller::addGeoreferencedImage(image_id imageID, const Georefer
...
+	}
+
+	// Generate upload data
+	auto uploadData = generateTileUploadData(cachedImageRecord->type, NDCToWorld, cachedImageRecord->georeferencedImageState.get());
```
Erfan-Ahmadi, line 919: It seems like it's caculating tiles to load and then requesting loading them.
maybe a better name along "computeTilesAndLoad"
I'm also thinking about if separating this into two function "compute" + "load" would do us any benefit when we use futures (instead of actually loading them on demand)
but I don't think so, this function can just return a bunch of futures to hold onto and wait upon...

62_CAD/DrawResourcesFiller.h
```diff
@@ -660,9 +692,9 @@ struct DrawResourcesFiller
 	 *
 	 * @param[out] outImageParams Structure to be filled with image creation parameters (format, size, etc.).
 	 * @param[out] outImageType Indicates whether the image should be fully resident or streamed.
-	 * @param[in] georeferencedImageParams Parameters describing the full image extents, viewport extents, and format.
+	 * @param[in] manager Manager for the georeferenced image
```
Erfan-Ahmadi, line 695: comment should probably change here

62_CAD/main.cpp
Fletterio: I didn't know the `.exr` could have all the mips by itself by default, so rn each mip level is a different `.exr`

62_CAD/Images.h
```diff
@@ -109,6 +109,71 @@ struct ImageCleanup : public core::IReferenceCounted
...
+
+		// 2. Change of Basis. Since {dirU, dirV} are orthogonal, the matrix to change from world coords to `span{dirU, dirV}` coords has a quite nice expression
+		//    Non-uniform scaling doesn't affect this, but this has to change if we allow for shearing (basis vectors stop being orthogonal)
+		float64_t2 dirU = retVal->georeferencedImageParams.worldspaceOBB.dirU;
+		float64_t2 dirV = float32_t2(dirU.y, -dirU.x) * retVal->georeferencedImageParams.worldspaceOBB.aspectRatio;
+		float64_t dirULengthSquared = nbl::hlsl::dot(dirU, dirU);
+		float64_t dirVLengthSquared = nbl::hlsl::dot(dirV, dirV);
+		float64_t2 firstRow = dirU / dirULengthSquared;
+		float64_t2 secondRow = dirV / dirVLengthSquared;
+		float64_t2x2 changeOfBasisMatrix(firstRow, secondRow);
```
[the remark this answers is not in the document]
Fletterio: Almost, the shift still needs to be post-rotated, so it's 2x2 times 2x1 product

62_CAD/Images.h
```diff
@@ -109,6 +109,71 @@ struct ImageCleanup : public core::IReferenceCounted
...
+	}
+
+	GeoreferencedImageParams georeferencedImageParams = {};
+	std::vector<std::vector<bool>> currentMappedRegionOccupancy = {};
+
+	// These are NOT UV, pixel or tile coords into the mapped image region, rather into the real, huge image
+	// Tile coords are always in mip 0 tile size. Translating to other mips levels is trivial
+	float64_t2 transformWorldCoordsToUV(const float64_t3 worldCoordsH) const { return nbl::hlsl::mul(world2UV, worldCoordsH); }
+	float64_t2 transformWorldCoordsToPixelCoords(const float64_t3 worldCoordsH) const { return float64_t2(georeferencedImageParams.imageExtents) * transformWorldCoordsToUV(worldCoordsH); }
+	float64_t2 transformWorldCoordsToTileCoords(const float64_t3 worldCoordsH, const uint32_t TileSize) const { return (1.0 / TileSize) * transformWorldCoordsToPixelCoords(worldCoordsH); }
```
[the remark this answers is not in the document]
Fletterio: Homogenous! To indicate what the third component is. If it's clear from context I can drop the H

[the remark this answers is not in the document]
Erfan-Ahmadi: Drop the H, no homo

62_CAD/DrawResourcesFiller.h
```diff
@@ -120,6 +122,36 @@ struct DrawResourcesFiller
...
+	asset::E_FORMAT queryGeoreferencedImageFormat(std::filesystem::path imagePath)
+	{
+		return georeferencedImageLoader->getFormat(imagePath);
+	}
+
+	// These are vulkan standard, might be different in n4ce!
+	constexpr static float64_t3 topLeftViewportNDC = float64_t3(-1.0, -1.0, 1.0);
+	constexpr static float64_t3 topRightViewportNDC = float64_t3(1.0, -1.0, 1.0);
+	constexpr static float64_t3 bottomLeftViewportNDC = float64_t3(-1.0, 1.0, 1.0);
+	constexpr static float64_t3 bottomRightViewportNDC = float64_t3(1.0, 1.0, 1.0);
```
Erfan-Ahmadi, line 154: I like that you documented your assumptions, we always use vulkan standard.

we will probably need to document these assumptions with a more "Global" settings.

After giving it some thought, I think the ndc standard choice should be documented for Nabla. (yes we use Vulkan internally) but  I think it's Nabla's job to decide and force a standard.

So my suggestion is move these locally to the function you're using them in and write a single comment about this assumption. It's no biggie

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -2649,4 +2684,256 @@ void DrawResourcesFiller::flushDrawObjects()
 		drawCalls.push_back(drawCall);
 		drawObjectsFlushedToDrawCalls = resourcesCollection.drawObjects.getCount();
 	}
+}
+
+DrawResourcesFiller::TileUploadData DrawResourcesFiller::generateTileUploadData(const ImageType imageType, const float64_t3x3& NDCToWorld, GeoreferencedImageStreamingState* imageStreamingState)
+{
+	// I think eventually it's better to just transform georeferenced images that aren't big enough into static images and forget about them
```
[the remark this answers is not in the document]
Erfan-Ahmadi: the criteria should be: the memory the image consumes is smaller than the GPU-resident image used for streaming

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -2649,4 +2684,256 @@ void DrawResourcesFiller::flushDrawObjects()
...
+
+	// Dangerous code - assumes image can be perfectly covered with tiles. Otherwise will need to handle edge cases
+	// TODO: All of this code only works for mip 0. Needs to be changed next to upload mip 1.
+	// Eventually this is all replaced by a few uploads to staging buffer + CS mip calc
+	for (uint32_t tileX = viewportTileRange.topLeft.x; tileX <= viewportTileRange.bottomRight.x; tileX++)
+	{
+		for (uint32_t tileY = viewportTileRange.topLeft.y; tileY <= viewportTileRange.bottomRight.y; tileY++)
+		{
+			// Compute tile offset relative to `currentMappedRegion.topLeft`, to get tile index into the gpu image
+			uint32_t2 gpuImageTileIndex = ((uint32_t2(tileX, tileY) - imageStreamingState->currentMappedRegion.topLeft) + imageStreamingState->gpuImageTopLeft) % imageStreamingState->gpuImageSideLengthTiles;
```
Erfan-Ahmadi, line 2803: I would make this a function in the imageStreamingState. It gets global tile index and returns gpuTileIndex, it's also and implicit documentation on how the tiling with toroidal updating works

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -2649,4 +2684,256 @@ void DrawResourcesFiller::flushDrawObjects()
...
+			uint32_t2 imageOffset = gpuImageTileIndex * GeoreferencedImageTileSize;
+			bufCopy.imageOffset = { imageOffset.x, imageOffset.y, 0u };
+			bufCopy.imageExtent.width = GeoreferencedImageTileSize;
+			bufCopy.imageExtent.height = GeoreferencedImageTileSize;
+			bufCopy.imageExtent.depth = 1;
+
+			tiles.emplace_back(imageStreamingState->georeferencedImageParams.format, std::move(tile), std::move(bufCopy));
+
+			// Upload the smaller tile to mip 1
+			tile = georeferencedImageLoader->load(imageStreamingState->georeferencedImageParams.storagePath, uint32_t2(tileX, tileY) * (GeoreferencedImageTileSize >> 1), uint32_t2(GeoreferencedImageTileSize, GeoreferencedImageTileSize) >> 1u, imageStreamingState->currentMappedRegion.baseMipLevel + 1);
```
Erfan-Ahmadi, line 2828: for `GeoreferencedImageTileSize >> 1` make another constexpr next to it called (GeoreferencedImageTileSizeMip1), it shows how we're using two mip levels and the tile size for loading mip 1 is different

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -2649,4 +2684,256 @@ void DrawResourcesFiller::flushDrawObjects()
...
+
+			asset::IImage::SBufferCopy bufCopy;
+			bufCopy.bufferOffset = 0;
+			bufCopy.bufferRowLength = GeoreferencedImageTileSize;
+			bufCopy.bufferImageHeight = 0;
+			bufCopy.imageSubresource.aspectMask = IImage::EAF_COLOR_BIT;
+			bufCopy.imageSubresource.mipLevel = 0u;
+			bufCopy.imageSubresource.baseArrayLayer = 0u;
+			bufCopy.imageSubresource.layerCount = 1u;
+			uint32_t2 imageOffset = gpuImageTileIndex * GeoreferencedImageTileSize;
```
Erfan-Ahmadi, line 2819: name gpuImageOffset for clarity that this is not related to the large georef image but the resident one

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -2649,4 +2684,256 @@ void DrawResourcesFiller::flushDrawObjects()
...
+	// Compute the mip level and tile range we would need to encompass the viewport
+	GeoreferencedImageTileRange viewportTileRange = computeViewportTileRange(NDCToWorld, imageStreamingState);
+
+	// A base mip level of x in the current mapped region means we can handle the viewport having mip level y, with x <= y < x + 1.0
+	// without needing to remap the region. When the user starts zooming in or out and the mip level of the viewport falls outside this range, we have to remap
+	// the mapped region.
+	const bool mipBoundaryCrossed = viewportTileRange.baseMipLevel != imageStreamingState->currentMappedRegion.baseMipLevel;
+
+	// If we moved a huge amount in any direction, no tiles will remain resident, so we simply reset state
+	const bool relativeShiftTooBig = nbl::hlsl::any
```
Erfan-Ahmadi, line 2704: add "this is considering no mip boundary crossing" to the comments of this, because it should be only evaluated when no mip boundary was crossed

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -2649,4 +2684,256 @@ void DrawResourcesFiller::flushDrawObjects()
...
+
+	// Last, we need to figure out an obb that covers only the currently loaded tiles
+
+	OrientedBoundingBox2D viewportWorldspaceOBB = imageStreamingState->georeferencedImageParams.worldspaceOBB;
+	// The original image `dirU` corresponds to `maxImageTileIndices.x + 1` mip 0 tiles (provided it's exactly that length in tiles)
+	// Dividing dirU by `maxImageTileIndices + (1,1)` we therefore get a vector that spans exactly one mip 0 tile (in the u direction) in worldspace. 
+	// Multiplying that by `2^mipLevel` we get a vector that spans exactly one mip `mipLevel` tile (in the u direction)
+	const float32_t2 oneTileDirU = imageStreamingState->georeferencedImageParams.worldspaceOBB.dirU / float32_t(imageStreamingState->fullImageLastTileIndices.x + 1u) * float32_t(1u << imageStreamingState->currentMappedRegion.baseMipLevel);
+	const float32_t2 oneTileDirV = float32_t2(oneTileDirU.y, -oneTileDirU.x) * viewportWorldspaceOBB.aspectRatio;
+	viewportWorldspaceOBB.topLeft += oneTileDirU * float32_t(viewportTileRange.topLeft.x);
```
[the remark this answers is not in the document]
Erfan-Ahmadi: make sure you document next to the struct that this value (topLeftTile) is based on the baseMipLevel

62_CAD/DrawResourcesFiller.cpp
Fletterio: On the first pass I just check how many tiles I need to upload, then reserve a vector of that length, then fill it out

Fletterio: But I think I get what you mean, I can return two pairs of uints from this first function and use that to get tiles from the loader and upload them

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -2649,4 +2684,256 @@ void DrawResourcesFiller::flushDrawObjects()
...
+	viewportWorldspaceOBB.topLeft += oneTileDirV * float32_t(viewportTileRange.topLeft.y);
+
+	const uint32_t2 viewportTileLength = viewportTileRange.bottomRight - viewportTileRange.topLeft + uint32_t2(1, 1);
+	viewportWorldspaceOBB.dirU = oneTileDirU * float32_t(viewportTileLength.x);
+	viewportWorldspaceOBB.aspectRatio = float32_t(viewportTileLength.y) / float32_t(viewportTileLength.x);
+	
+	// UV logic currently ONLY works when the image not only fits an integer amount of tiles, but also when it's a PoT amount of them
+	// (this means every mip level also gets an integer amount of tiles).
+	// When porting to n4ce, for the image to fit an integer amount of tiles (instead of rewriting the logic) we can just pad the right/bottom sides with alpha=0 pixels
```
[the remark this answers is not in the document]
Fletterio: Ah wait nvm I can just clamp the uv ig? I'll give that a try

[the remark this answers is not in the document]
Erfan-Ahmadi: nice catch! so even if we copy only a sub-region the other parts should be invisible and not hold some previous data.
easy fix for now would be to return full tile from the loader and set alpha to 0.
Playing with UV is also a smart idea, will work. note that this issue will only pop-up at the right or bottom side of the larger sized image where you "shrink" your mapped region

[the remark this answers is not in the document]
Erfan-Ahmadi: You seem to know what to do. Do it and show me how it works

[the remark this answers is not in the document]
Erfan-Ahmadi: ouch... 
so many problems with nPoT (discussions in discord since a few days ago)

PR #207, opened 2025-08-05
--savemesh argument in example 12

YasInvolved:


12_MeshLoaders/main.cpp
```diff
@@ -176,6 +195,32 @@ class MeshLoadersApp final : public MonoWindowApplication, public BuiltinResourc
...
+				if (!std::filesystem::exists(prefix))
+					m_system->createDirectory(prefix);
+
+				auto savePath = (prefix / path(m_modelPath).filename()).generic_string();
+			
+				m_logger->log("Saving mesh to %S", ILogger::ELL_INFO, savePath.c_str()); 
+				// TODO (Yas): learn how to get out the geometry from renderer and transform it into IAsset
+				
+				auto& asset = m_currentBundle.getContents()[0];
+				IAssetWriter::SAssetWriteParams params{ asset.get() };
```
devshgraphicsprogramming, line 214: only what you're intending to write shall be kept

12_MeshLoaders/main.cpp
```diff
@@ -416,6 +461,11 @@ class MeshLoadersApp final : public MonoWindowApplication, public BuiltinResourc
 		Camera camera = Camera(core::vectorSIMDf(0, 0, 0), core::vectorSIMDf(0, 0, 0), core::matrix4SIMD());
 		// mutables
 		std::string m_modelPath;
+
+		SAssetBundle m_currentBundle;
```
devshgraphicsprogramming, line 465: grab only the geometry, see https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/207/files#r2266047038

12_MeshLoaders/main.cpp
```diff
@@ -1,6 +1,7 @@
 // Copyright (C) 2018-2020 - DevSH Graphics Programming Sp. z O.O.
 // This file is part of the "Nabla Engine".
 // For conditions of distribution and use, see copyright notice in nabla.h
+#include "../3rdparty/argparse/include/argparse/argparse.hpp"
```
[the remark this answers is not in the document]
AnastaZIuk:
```c++
#include "argparse/argparse.hpp"
```

```cmake
add_dependencies(${EXECUTABLE_NAME} argparse)
target_include_directories(${EXECUTABLE_NAME} PUBLIC $<TARGET_PROPERTY:argparse,INTERFACE_INCLUDE_DIRECTORIES>)
```

@YasInvolved don't use relative paths use the ones from include search dirs

12_MeshLoaders/main.cpp
```diff
@@ -446,6 +438,24 @@ class MeshLoadersApp final : public MonoWindowApplication, public BuiltinResourc
 			return true;
 		}
 
+		void writeGeometry()
+		{
+			// make save path
+			static const auto prefix = std::filesystem::absolute("saved/");
+
+			if (!std::filesystem::exists(prefix))
+				m_system->createDirectory(prefix);
```
devshgraphicsprogramming, line 447: path is wrong, see the 4 CWD members of the apocalypse in the basic framework app

12_MeshLoaders/main.cpp
```diff
@@ -197,22 +197,9 @@ class MeshLoadersApp final : public MonoWindowApplication, public BuiltinResourc
 
 		inline bool onAppTerminated() override
 		{
-			if (m_saveMeshOnExit)
+			if (m_saveGeomOnExit)
```
devshgraphicsprogramming, line 200: check for current geom

12_MeshLoaders/main.cpp
```diff
@@ -446,6 +438,24 @@ class MeshLoadersApp final : public MonoWindowApplication, public BuiltinResourc
...
+				m_system->createDirectory(prefix);
+
+			auto savePath = (prefix / path(m_modelPath).filename()).generic_string();
+
+			m_logger->log("Saving mesh to %S", ILogger::ELL_INFO, savePath.c_str());
+
+			// should I do a const cast here?
+			const IAsset* asset = m_currentGeom.get();
+			IAssetWriter::SAssetWriteParams params{ const_cast<IAsset*>(asset) };
+			m_assetMgr->writeAsset(savePath, params);
```
devshgraphicsprogramming, line 456: after writing out once, the `m_currentGeom needs to be nulled out

12_MeshLoaders/main.cpp
```diff
@@ -305,9 +321,12 @@ class MeshLoadersApp final : public MonoWindowApplication, public BuiltinResourc
...
 			// TODO: do it async
 			if (m_saveGeom)
-				writeGeometry(
-					const_cast<ICPUPolygonGeometry*>(geometries[0].get()), 
-					m_specifiedGeomSavePath.value_or((m_saveGeomPrefixPath / path(m_modelPath).filename()).generic_string())
+				m_saveGeomTaskFuture = std::async(
+					std::launch::async,
+					[this, geometries] { writeGeometry(
+						geometries[0],
+						m_specifiedGeomSavePath.value_or((m_saveGeomPrefixPath / path(m_modelPath).filename()).generic_string())
```
devshgraphicsprogramming, line 328: data race hazard, `m_specifiedGeomSavePath` can change while old async is still running if the mesh writing is slow enough

make a temp copy of the variable and capture the copy by value

Also await the previous already running async if there is one because move-replacing with a new async

YasInvolved: We could schedule geometry saving tasks by using a queue and then await them all at application exit.

PR #209, opened 2025-08-11
bitonic sort sample 

CrabExtra:


devshgraphicsprogramming:
@Fletterio wanna try your hand at a review?

CMakeLists.txt
```diff
@@ -96,19 +97,9 @@ if(NBL_BUILD_EXAMPLES)
...
-
-			if(NBL_EMBED_BUILTIN_RESOURCES)
-				LINK_BUILTIN_RESOURCES_TO_TARGET(${T} NblExtExamplesAPIBuiltinsSource)
-				LINK_BUILTIN_RESOURCES_TO_TARGET(${T} NblExtExamplesAPIBuiltinsInclude)
-				LINK_BUILTIN_RESOURCES_TO_TARGET(${T} NblExtExamplesAPIBuiltinsBuild)
-			endif()
-		endif()
+        target_link_libraries(${T} PUBLIC ${NBL_EXAMPLES_API_TARGET})
+		target_include_directories(${T} PUBLIC $<TARGET_PROPERTY:${NBL_EXAMPLES_API_TARGET},INCLUDE_DIRECTORIES>)
+		target_precompile_headers(${T} REUSE_FROM "${NBL_EXAMPLES_API_TARGET}")
```
devshgraphicsprogramming, line 102: this needs to match `master`, no difference

13_BitonicSort/app_resources/bitonic_sort_shader.comp.hlsl
```diff
@@ -0,0 +1,112 @@
+#include "nbl/builtin/hlsl/bda/bda_accessor.hlsl"
+
+struct BitonicPushData
+{
+    uint64_t inputKeyAddress;
+    uint64_t inputValueAddress;
+    uint64_t outputKeyAddress;
+    uint64_t outputValueAddress;
+    uint32_t dataElementCount;
+};
```
devshgraphicsprogramming, line 10: why are you not including `common.hlsl` and having duplicate code?

13_BitonicSort/app_resources/bitonic_sort_shader.comp.hlsl
```diff
@@ -0,0 +1,112 @@
...
+
+using namespace nbl::hlsl;
+
+[[vk::push_constant]] BitonicPushData pushData;
+
+using DataPtr = bda::__ptr<uint32_t>;
+using DataAccessor = BdaAccessor<uint32_t>;
+
+groupshared uint32_t sharedKeys[ElementCount];
```
devshgraphicsprogramming, line 19: Good SoA instead of AoS, no bank conflicts

13_BitonicSort/app_resources/bitonic_sort_shader.comp.hlsl
```diff
@@ -0,0 +1,112 @@
...
+    const uint32_t threadId = localId.x;
+    const uint32_t dataSize = pushData.dataElementCount;
+    
+    DataAccessor inputKeys = DataAccessor::create(DataPtr::create(pushData.inputKeyAddress));
+    DataAccessor inputValues = DataAccessor::create(DataPtr::create(pushData.inputValueAddress));
+    
+    for (uint32_t i = threadId; i < dataSize; i += WorkgroupSize)
+    {
+        inputKeys.get(i, sharedKeys[i]);
+        inputValues.get(i, sharedValues[i]);
```
devshgraphicsprogramming, line 35: Very nice but why isn't this wrapped up into a statically polymorphic struct like FFT and workgroup prefix sum? So we can actually reuse the code ? You know with shared memory accessors etc.

13_BitonicSort/app_resources/bitonic_sort_shader.comp.hlsl
```diff
@@ -0,0 +1,112 @@
...
+    }
+    
+    // Synchronize all threads after loading
+    GroupMemoryBarrierWithGroupSync();
+    
+
+    for (uint32_t stage = 0; stage < Log2ElementCount; stage++)
```
devshgraphicsprogramming, line 42: Where does Log2ElementCount come from? I have a feeling that you shouldn't have a data size push constant then.

13_BitonicSort/main.cpp
```diff
@@ -0,0 +1,323 @@
...
+	{
+		if (!device_base_t::onAppInitialized(smart_refctd_ptr(system)))
+			return false;
+		if (!asset_base_t::onAppInitialized(std::move(system)))
+			return false;
+
+		auto limits = m_physicalDevice->getLimits();
+		const uint32_t max_shared_memory_size = limits.maxComputeSharedMemorySize;
+		const uint32_t max_workgroup_size = limits.maxComputeWorkGroupInvocations; // Get actual GPU limit
+		const uint32_t bytes_per_elements = sizeof(uint32_t) * 2; // 2 uint32_t per element (key and value)
```
devshgraphicsprogramming, line 36: you should get this with `sizeof()` of the structs in `app_resources/common.hlsl`

13_BitonicSort/main.cpp
```diff
@@ -0,0 +1,323 @@
...
+		const uint32_t max_element_in_shared_memory = max_shared_memory_size / bytes_per_elements;
```
devshgraphicsprogramming, line 37: if you use all the shared memory, you won't have room for as many workgroups as possible.

Btw a good undocumented rule for occupancy is to make each invocation use no more than 32 bytes of shared memory.

Also all these calculations can be done with templates in HLSL/C++ (see the workgroup FFT and Prefix sum code)

13_BitonicSort/main.cpp
```diff
@@ -0,0 +1,323 @@
...
+				buffer = m_device->createBuffer(std::move(params));
+				if (!buffer)
+					logFail("Failed to create GPU buffer of size %d!\n", buffer_size);
+
+				buffer->setObjectDebugName(label);
+
+				auto reqs = buffer->getMemoryReqs();
+				reqs.memoryTypeBits &= m_physicalDevice->getHostVisibleMemoryTypeBits();
+
+				*allocation = m_device->allocate(reqs, buffer.get(), IDeviceMemoryAllocation::EMAF_DEVICE_ADDRESS_BIT);
```
devshgraphicsprogramming, line 123: nice and clean

13_BitonicSort/main.cpp
```diff
@@ -0,0 +1,323 @@
...
+			buffers[2]->getDeviceAddress(),
+			buffers[3]->getDeviceAddress()
+		};
+
+	
+		void* mapped_memory[] = {
+			allocation[0].memory->map({0ull,allocation[0].memory->getAllocationSize()}, IDeviceMemoryAllocation::EMCAF_READ),
+			allocation[1].memory->map({0ull,allocation[1].memory->getAllocationSize()}, IDeviceMemoryAllocation::EMCAF_READ),
+			allocation[2].memory->map({0ull,allocation[2].memory->getAllocationSize()}, IDeviceMemoryAllocation::EMCAF_READ),
+			allocation[3].memory->map({0ull,allocation[3].memory->getAllocationSize()}, IDeviceMemoryAllocation::EMCAF_READ)
```
devshgraphicsprogramming, line 147: Vulkan doesn't care about the EMCAF flags, but they're kinda wrong here, I'm pretty sure you WRITE into them with the host, while others you READ at the end

13_BitonicSort/main.cpp
```diff
@@ -0,0 +1,323 @@
...
+		if (!mapped_memory[0] || !mapped_memory[1] || !mapped_memory[2] || !mapped_memory[3])
+			return logFail("Failed to map the Device Memory!\n");
+
+		// Generate random data
+		unsigned seed = std::chrono::system_clock::now().time_since_epoch().count();
+		std::mt19937 g(seed);
+
+		auto bufferData = new uint32_t * [2];
+		for (int i = 0; i < 2; ++i) {
+			bufferData[i] = new uint32_t[element_count];
```
devshgraphicsprogramming, line 158: use a `std::array<std::unique_ptr<uint32_t>,2>`

13_BitonicSort/app_resources/bitonic_sort_shader.comp.hlsl
```diff
@@ -0,0 +1,112 @@
...
+
+using namespace nbl::hlsl;
+
+[[vk::push_constant]] BitonicPushData pushData;
+
+using DataPtr = bda::__ptr<uint32_t>;
+using DataAccessor = BdaAccessor<uint32_t>;
+
+groupshared uint32_t sharedKeys[ElementCount];
+groupshared uint32_t sharedValues[ElementCount];
```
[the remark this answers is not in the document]
devshgraphicsprogramming: also because the value never has to get accessed until the end.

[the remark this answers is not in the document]
devshgraphicsprogramming: hmm actually you wouldn't initialize `sharedValueIndex` with IOTA but the result of the `subgroup::BitonicSort`

13_BitonicSort/app_resources/bitonic_sort_shader.comp.hlsl
```diff
@@ -0,0 +1,112 @@
...
+    
+
+    for (uint32_t stage = 0; stage < Log2ElementCount; stage++)
+    {
+        for (uint32_t pass = 0; pass <= stage; pass++)
+        {
+            const uint32_t compareDistance = 1 << (stage - pass);
+            
+            for (uint32_t i = threadId; i < dataSize; i += WorkgroupSize)
+            {
```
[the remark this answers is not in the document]
devshgraphicsprogramming: one slight nitpick, this won't `[unroll]` were the `dataSize` a compile time constant.

to get a loop that unrolls you need to do
```cpp
for (uint32_t baseIndex=0; baseIndex<dataSize; baseIndex+=WorkgroupSize)
{
   const uint32_t i = threadId+baseIndex;
```
compilers are dumb

13_BitonicSort/app_resources/bitonic_sort_shader.comp.hlsl
```diff
@@ -0,0 +1,112 @@
...
+
+groupshared uint32_t sharedKeys[ElementCount];
+groupshared uint32_t sharedValues[ElementCount];
+
+[numthreads(WorkgroupSize, 1, 1)]
+[shader("compute")]
+void main(uint32_t3 dispatchId : SV_DispatchThreadID, uint32_t3 localId : SV_GroupThreadID)
+{
+    const uint32_t threadId = localId.x;
+    const uint32_t dataSize = pushData.dataElementCount;
```
devshgraphicsprogramming, line 27: I'd make that constant and part of the Config, because the workgroup size needs to be hardcoded and as I've mentioned optimal amount of shared memory per invocation to use is 8 dwords (32 bytes)

13_BitonicSort/app_resources/bitonic_sort_shader.comp.hlsl
```diff
@@ -0,0 +1,112 @@
...
+                }
+            }
+            
+            GroupMemoryBarrierWithGroupSync();
+        }
+    }
+    
+
+    DataAccessor outputKeys = DataAccessor::create(DataPtr::create(pushData.outputKeyAddress));
+    DataAccessor outputValues = DataAccessor::create(DataPtr::create(pushData.outputValueAddress));
```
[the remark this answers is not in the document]
devshgraphicsprogramming: Items Per Invocation should also come from Config.

Btw with a uint16_t value index and 1 invocation always processing a pair of items, this means a baseline shared memory consumption of 8 bytes per invocation if the key is also either 2 bytes or an index of a key.

Realistically max virtual thread repetition is 4, in your case with an actual 32bit key getting used directly, its 3 but with optimal workgroup size of 512 on NV to keep the multiple PoT it becomes 2.

As you can see the point at which one should consider indexing keys instead of using them out right is 14 byte keys.

13_BitonicSort/app_resources/bitonic_sort_shader.comp.hlsl
```diff
@@ -0,0 +1,112 @@
...
+                    partnerValue = WaveReadLaneAt(myValue, partnerLane);
+                }
+                else
+                {
+                    // SHARED MEM
+                    myKey = sharedKeys[i];
+                    myValue = sharedValues[i];
+                    partnerKey = sharedKeys[partnerId];
+                    partnerValue = sharedValues[partnerId];
+                }
```
devshgraphicsprogramming, line 79: the wholly subgroup stages (stage<=subgroupSizeLog2) need separation from the stages where the compareDistance can be big. Thats what I want wrapped up in `subgroup::BitonicSort`

13_BitonicSort/app_resources/bitonic_sort_shader.comp.hlsl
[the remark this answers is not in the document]
devshgraphicsprogramming: btw when doing subgroup bitonics its also possible to have multiple items per invocation (more than 2), so its useful to bitonic sort in registers first (see subgroup prefix sum and FFT doing stuff locally within an invocation first).

13_BitonicSort/app_resources/bitonic_sort_shader.comp.hlsl
devshgraphicsprogramming: btw the workgroup stages would still use subgroup to finish off, because `compareDistance` decreases eventually

13_BitonicSort/app_resources/bitonic_sort_shader.comp.hlsl
```diff
@@ -0,0 +1,112 @@
...
+
+                const uint32_t sequenceSize = 1 << (stage + 1);
+                const uint32_t sequenceIndex = i / sequenceSize;
+                const bool sequenceAscending = (sequenceIndex % 2) == 0;
+                const bool ascending = true;
+                const bool finalDirection = sequenceAscending == ascending;
+                
+                const bool swap = (myKey > partnerKey) == finalDirection;
```
devshgraphicsprogramming, line 87: this can be simply expressed as `bool(i & (0x1u<< (stage+2)))`, all that code is just checking for one bit being set in `i`

13_BitonicSort/app_resources/bitonic_sort_shader.comp.hlsl
```diff
@@ -0,0 +1,112 @@
...
+                
+                // WORKGROUP COORDINATION: Only lower-indexed element writes both
+                if (i < partnerId && swap)
+                {
+                    sharedKeys[i] = partnerKey;
+                    sharedKeys[partnerId] = myKey;
+                    sharedValues[i] = partnerValue;
+                    sharedValues[partnerId] = myValue;
+                }
```
devshgraphicsprogramming, line 96: the key is not using 2 invocations for 2 inputs, but 1 for 2, throughout the function

13_BitonicSort/app_resources/bitonic_sort_shader.comp.hlsl
```diff
@@ -0,0 +1,112 @@
...
+        for (uint32_t pass = 0; pass <= stage; pass++)
+        {
+            const uint32_t compareDistance = 1 << (stage - pass);
+            
+            for (uint32_t i = threadId; i < dataSize; i += WorkgroupSize)
+            {
+                const uint32_t partnerId = i ^ compareDistance;
+                
+                if (partnerId >= dataSize)
+                    continue;
```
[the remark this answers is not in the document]
devshgraphicsprogramming: btw you might have some issues with making this work for a single subgroup, @Fletterio may have some ideas how to help you there (see his FFT blogpost, about trading halves of FFT with subgroup shuffles while only keeping one invocation active).

The main problem is that with a 32-sized subgroup you need to treat 64 items.

Ideally you'd want one invocation to get 2 keys and value indices, but keep them rearranged in such a way that always 32 invocations are active.

Meaning that with step sizes of 1,2,4,8,... subgroupSize/2 you need to cleverly "reuse" the "partner ID" invocation to do the work of the "would be" second subgroup.

13_BitonicSort/main.cpp
```diff
@@ -0,0 +1,323 @@
...
+			m_api->startCapture();
+			queue->submit(submit_infos);
+			m_api->endCapture();
+		}
+
+		const ISemaphore::SWaitInfo wait_infos[] = { {
+				.semaphore = progress.get(),
+				.value = timeline
+			} };
+		m_device->blockForSemaphores(wait_infos);
```
devshgraphicsprogramming, line 249: add a separate benchmark mode that does more workgroup dispatches (but allocates the buffer on DeviceLocal without redability) and on a loop with a time elapsed query

If you're curious you can see if you're faster than https://github.com/AndrewBoessen/Bitonic-Merge-Sort

Fletterio (commented):
Just realized these were all pending, most are outdated see if any still hold up

13_BitonicSort/app_resources/common.hlsl
```diff
@@ -0,0 +1,13 @@
+#ifndef _BITONIC_SORT_COMMON_INCLUDED_
+#define _BITONIC_SORT_COMMON_INCLUDED_
+#include "nbl/builtin/hlsl/cpp_compat.hlsl"
+
+struct PushConstantData
+{
+	uint64_t deviceBufferAddress;
+};
+
+NBL_CONSTEXPR uint32_t WorkgroupSizeLog2 = 10;  // 1024 threads (2^10)
```
devshgraphicsprogramming: 512 is optimal residency on all GPUs

13_BitonicSort/main.cpp
```diff
@@ -0,0 +1,319 @@
...
+			// Generate random input data
+			unsigned seed = std::chrono::system_clock::now().time_since_epoch().count();
+			std::mt19937 g(seed);
+
+			std::cout << "Input array: ";
+			for (uint32_t i = 0; i < elementCount; i++) {
+				uint32_t key = g() % 10000;
+				uint32_t value = i; // Use index as value for stable sorting
+				inputPtr[i * 2] = key;
+				inputPtr[i * 2 + 1] = value;
```
devshgraphicsprogramming: again another reason to have a named struct so you're not doing this unreadable type punning

13_BitonicSort/main.cpp
```diff
@@ -0,0 +1,319 @@
...
+				std::cout << "(" << key << "," << value << "), ";
+				if ((i + 1) % 20 == 0) {
+					std::cout << "\n";
+				}
+			}
+			std::cout << "\nElement count: " << elementCount << "\n";
```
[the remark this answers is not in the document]
devshgraphicsprogramming: or an `std::ostringstream` and then `m_logger` to print the contents

13_BitonicSort/main.cpp
```diff
@@ -0,0 +1,319 @@
...
+			barrier1.range.buffer = m_deviceLocalBuffer;
+			barrier1.barrier.dep.srcStageMask = PIPELINE_STAGE_FLAGS::COPY_BIT;
+			barrier1.barrier.dep.srcAccessMask = ACCESS_FLAGS::MEMORY_WRITE_BITS;
+			barrier1.barrier.dep.dstStageMask = PIPELINE_STAGE_FLAGS::COMPUTE_SHADER_BIT;
+			barrier1.barrier.dep.dstAccessMask = ACCESS_FLAGS::MEMORY_READ_BITS;
+			cmdbuf->pipelineBarrier(asset::E_DEPENDENCY_FLAGS(0), pipelineBarrierInfo1);
+
+			cmdbuf->pushConstants(m_pipeline->getLayout(), IShader::E_SHADER_STAGE::ESS_COMPUTE, 0u, sizeof(pc), &pc);
+
+			cmdbuf->dispatch(1, 1, 1);
```
devshgraphicsprogramming: I'd consider preparing multiple test data sets, right now you only have one random one, and having 1 workgroup per test (so still one dispatch).

I'd add:
- all keys equal
- keys already sorted
- keys in exact reverse
as manual test cases.

Furthermore this should be a stable sort (unlike the a counting sort), its important to check things like stability so making some test with some neighbouring equal keys in the input array is important to check they have not changed places.

13_BitonicSort/main.cpp
```diff
@@ -0,0 +1,319 @@
...
+				bool is_sorted = true;
+				int32_t error_index = -1;
+				for (uint32_t i = 1; i < elementCount; i++) {
+					uint32_t prevKey = data[(i - 1) * 2];
+					uint32_t currKey = data[i * 2];
+					if (currKey < prevKey) {
+						is_sorted = false;
+						error_index = i;
+						break;
+					}
```
devshgraphicsprogramming: because Bitonic _should_ be stable, you should also check 
```cpp
else if (currKey==prevKey && currValue>prevValue) // check stability, this is why we've initialized the values in such a particular way
{
   // fail, not to bad sorting but instability
}
```

and also `assert(curValue!=prevValue);` because of how we initialized the values

CMakeLists.txt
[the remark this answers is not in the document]
devshgraphicsprogramming: now it is

PR #238, opened 2025-12-24
Image upload benchamark

CrabExtra:
Added simple Image uploading benchmark


73_ImageUploadBenchmark/main.cpp
```diff
@@ -0,0 +1,443 @@
...
+			m_logger->log("DEVICE_LOCAL memory types not found!", ILogger::ELL_ERROR);
+			return false;
+		}
+
+		IQueue* queue = getQueue(IQueue::FAMILY_FLAGS::GRAPHICS_BIT);
+		smart_refctd_ptr<IGPUImage> destinationImage;
+		{
+			IGPUImage::SCreationParams imgParams{};
+			imgParams.type = IImage::E_TYPE::ET_2D;
+			imgParams.extent.width = TILE_SIZE * 32;
```
Erfan-Ahmadi: this should depend on your TILES_PER_FRAME.

73_ImageUploadBenchmark/main.cpp
```diff
@@ -0,0 +1,443 @@
...
+				};
+				m_device->blockForSemaphores({&frameWaitInfo, 1});
+			}
+			auto t2 = std::chrono::high_resolution_clock::now();
+
+			commandPools[cmdBufIndex]->reset();
+
+			uint32_t bufferOffset = cmdBufIndex * partitionSize;
+			void* targetPtr = static_cast<uint8_t*>(mappedPtr) + bufferOffset;
+			memcpy(targetPtr, cpuSourceData, partitionSize);
```
Erfan-Ahmadi: you need to check if the mapped pointer's buffer was created on HOST_COHERENT memory. if not, you need to flush the range to be available for the gpu

73_ImageUploadBenchmark/main.cpp
```diff
@@ -0,0 +1,443 @@
...
+			commandPools[i]->createCommandBuffers(
+				IGPUCommandPool::BUFFER_LEVEL::PRIMARY,
+				1,
+				&commandBuffers[i]
+			);
+		}
+
+		uint64_t timelineValue = 0;
+
+		commandBuffers[0]->begin(IGPUCommandBuffer::USAGE::ONE_TIME_SUBMIT_BIT);
```
Erfan-Ahmadi: It's ok to do this if the "end-purpose" of this image was to be a destination image and nothing else :D
but it's going to be used in the shaders as well. so what's the first thing you'll think of doing? transition to TRANSFER__DST for the copy and SHADER_READ_OPTIMAL after the copy. it makes sense for a single-upload image which you're going to use in GPU in many many future frames. but we're potentially uploading to this image every frame and it would be redudndant to do this image layout transitions each time.
that's why we asked you to keep the gpu image in GENERAL layout.

another note:
- pipeline barriers are not for image layout transitions only. you still need them to ensure correct synchronization and access to the resource. 
so: keep the image as GENERAL layout + add 2 pipeline barriers (1 before the transfer, 1 after) the one after is a "Fake" or "mock" one for now. it's as if it's going to be read by a shader soon. 
hint: the second barrier after the transfer should have `SHADER_READ_BIT ` as access in `FRAGMENT` stage (with no layout transition, i.e GENERAL->GENERAL)
- you'll most likely get a sync validation error ([assuming you turned it on](https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/238/files#r2646794459)) if you don't have any pipeline barrier for this destination image

73_ImageUploadBenchmark/main.cpp
```diff
@@ -0,0 +1,443 @@
...
+			commandPools[cmdBufIndex]->reset();
+
+			uint32_t bufferOffset = cmdBufIndex * partitionSize;
+			void* targetPtr = static_cast<uint8_t*>(mappedPtr) + bufferOffset;
+			memcpy(targetPtr, cpuSourceData, partitionSize);
+			auto t3 = std::chrono::high_resolution_clock::now();
+
+			commandBuffers[cmdBufIndex]->begin(IGPUCommandBuffer::USAGE::ONE_TIME_SUBMIT_BIT);
+			commandBuffers[cmdBufIndex]->copyBufferToImage(
+				stagingBuffer,
```
Erfan-Ahmadi: ok good, single copy call for all tiles. remember to change this to GENERAL after https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/238/files#r2646807113

73_ImageUploadBenchmark/main.cpp
```diff
@@ -0,0 +1,443 @@
...
+			.value = timelineValue
+		};
+		m_device->blockForSemaphores({&finalWait, 1});
+
+		auto endTime = std::chrono::high_resolution_clock::now();
+
+		delete[] cpuSourceData;
+		for (uint32_t i = 0; i < framesInFlight; i++)
+			delete[] regionsPerFrame[i];
+		delete[] regionsPerFrame;
```
Erfan-Ahmadi: I usually cringe when I see raw new/delete in the code. (PTSD from seeing legacy code)
depending on the situation just use:
- stack allocation type data[MAX_SIZE]
- vector for more variation and dyanmically sized stuff
- smart_refctd_ptr for nabla refcounted stuff (inherited from IRefCounted)
- unique_ptr for custom non nabla stuff
and take it easy. it's just an example/benchmark

73_ImageUploadBenchmark/app_resources/tile_upload.comp.hlsl
```diff
@@ -0,0 +1,260 @@
...
+    return x;
+}
+
+uint32_t2 mortonDecode(uint32_t code)
+{
+    return uint32_t2(
+        mortonCompact1By1(code),
+        mortonCompact1By1(code >> 1u)
+    );
+}
```
Erfan-Ahmadi: we already have morton code function in our hlsl library.
already mentioned here: https://discord.com/channels/593902898015109131/1450452175234011147/1471719738525614334

remove this and include "nbl/builtin/hlsl/morton.hlsl".
if usage is not clear search on discord and ask

73_ImageUploadBenchmark/main.cpp
```diff
@@ -0,0 +1,1572 @@
...
+		m_logger->log("    Record time: %.3f s (%.1f%%)", ILogger::ELL_PERFORMANCE, totalRecordTime, 100.0 * totalRecordTime / elapsedSeconds);
+		m_logger->log("    Submit time: %.3f s (%.1f%%)", ILogger::ELL_PERFORMANCE, totalSubmitTime, 100.0 * totalSubmitTime / elapsedSeconds);
+		double memcpyGBps = totalGB / totalMemcpyTime;
+		m_logger->log("    Memcpy speed: %.2f GB/s", ILogger::ELL_PERFORMANCE, memcpyGBps);
+
+		return { wallThroughputGBps, gpuThroughputGBps, memcpyGBps };
+	}
+
+
+	double runBenchmarkImageStaging(
```
Erfan-Ahmadi, line 739: this was a failed attempt because we realized we cannot work with preinitialized images and optimal layout in host visible memory, correct?

73_ImageUploadBenchmark/main.cpp
```diff
@@ -0,0 +1,1572 @@
...
+		}
+
+		m_logger->log("Setup complete. Running verification loop (%u frames)", ILogger::ELL_PERFORMANCE, VERIFICATION_LOOP_COUNT);
+
+		return true;
+	}
+
+	bool keepRunning() override { return m_frameIndex < VERIFICATION_LOOP_COUNT; }
+
+	void workLoopBody() override
```
Erfan-Ahmadi: what is this workLoopBody?
I expected it to be empty and benchmarks just running in app initialization and then app gets terminated

Erfan-Ahmadi: If you need to verify your compute algorithms, just run them once after the main benchmarks?
I don't get the point of `VERIFICATION_LOOP_COUNT`

Erfan-Ahmadi: but good attempt to verify your results. always good practice,

73_ImageUploadBenchmark/main.cpp
```diff
@@ -0,0 +1,1572 @@
...
+	IDeviceMemoryAllocator::SAllocation m_mortonReadbackAlloc;
+	void* m_stagingMappedPtr = nullptr;
+	void* m_snakeReadbackMappedPtr = nullptr;
+	void* m_mortonReadbackMappedPtr = nullptr;
+	smart_refctd_ptr<IGPUCommandPool> m_cmdPool;
+	smart_refctd_ptr<IGPUCommandBuffer> m_cmdbuf;
+	smart_refctd_ptr<ISemaphore> m_sem;
+	uint32_t m_frameIndex = 0;
+
+	void runAllBenchmarks()
```
Erfan-Ahmadi, line 335: this function looks good

73_ImageUploadBenchmark/main.cpp
```diff
@@ -0,0 +1,1572 @@
...
+	}
+
+	struct BenchResult
+	{
+		double wallGBps;
+		double gpuGBps;
+		double memcpyGBps;
+	};
+
+	void generateTileCopyRegions(
```
Erfan-Ahmadi, line 457: looks ok, I had already reviewed it

73_ImageUploadBenchmark/main.cpp
```diff
@@ -0,0 +1,1572 @@
...
+			outRegions[i].imageOffset = { tileX, tileY, 0 };
+			outRegions[i].imageExtent = { tileSize, tileSize, 1 };
+			outRegions[i].imageSubresource.aspectMask = IImage::E_ASPECT_FLAGS::EAF_COLOR_BIT;
+			outRegions[i].imageSubresource.mipLevel = 0;
+			outRegions[i].imageSubresource.baseArrayLayer = 0;
+			outRegions[i].imageSubresource.layerCount = 1;
+		}
+	}
+
+	BenchResult runBenchmark(
```
Erfan-Ahmadi: rename to runBenchmarkBufferToImageCopyCommand or something better that shows it's different than the compute one

73_ImageUploadBenchmark/main.cpp
```diff
@@ -0,0 +1,1572 @@
...
+
+		m_logger->log("  Timing breakdown for %s:", ILogger::ELL_PERFORMANCE, strategyName);
+		m_logger->log("    Wait time:   %.3f s (%.1f%%)", ILogger::ELL_PERFORMANCE, totalWaitTime, 100.0 * totalWaitTime / elapsedSeconds);
+		m_logger->log("    Memcpy time: %.3f s (%.1f%%)", ILogger::ELL_PERFORMANCE, totalMemcpyTime, 100.0 * totalMemcpyTime / elapsedSeconds);
+		m_logger->log("    Record time: %.3f s (%.1f%%)", ILogger::ELL_PERFORMANCE, totalRecordTime, 100.0 * totalRecordTime / elapsedSeconds);
+		m_logger->log("    Submit time: %.3f s (%.1f%%)", ILogger::ELL_PERFORMANCE, totalSubmitTime, 100.0 * totalSubmitTime / elapsedSeconds);
+		double memcpyGBps = totalGB / totalMemcpyTime;
+		m_logger->log("    Memcpy speed: %.2f GB/s", ILogger::ELL_PERFORMANCE, memcpyGBps);
+
+		return { wallThroughputGBps, gpuThroughputGBps, memcpyGBps };
```
Erfan-Ahmadi, line 735: this functions looks good and sane too

73_ImageUploadBenchmark/main.cpp
Erfan-Ahmadi: let's remove the body of the function and replace it with comment, save some LoC and add some documentation here on why we failed.

73_ImageUploadBenchmark/app_resources/tile_upload.comp.hlsl
```diff
@@ -0,0 +1,72 @@
+#include "common.hlsl"
+
+[[vk::binding(0,0)]] RWTexture2D<float32_t4> dstImage;
+[[vk::push_constant]] PushConstantData pc;
+
+using namespace nbl::hlsl;
+
+static const uint32_t TILE_SIZE = 128u;
+
+[numthreads(128, 1, 1)]
```
Erfan-Ahmadi: - try 2D workgroups.
- make workgroup size 128x4 (make sure it reflects on you dispatch) --> 128*4=512 is a good workgroup size, it can fit exactly 3 workgroups on modern SMs 
- ~~pixelPos will be just the global thread idx.~~ we need to have individual offsets for each tile requests
- your tile size is PoT, you could just use bitshift with 7 for division `<< TILE_SIZE_LOG2`. for modulo use `&127u` or `&(TILE_SIZE-1)`
- tileIdx will `globalPos.xy << 7` (define TILE_SIZE_LOG2)
- localPos will be globalPos.xy&127
- your start read location will be tileIdx*TILE_SIZE*TILE_SIZE or tileIdx << 14u
- you get it :D I won't continue

73_ImageUploadBenchmark/app_resources/tile_upload.comp.hlsl
```diff
@@ -0,0 +1,72 @@
...
+    const uint32_t2 tileBase = uint32_t2(tileIdx % pc.tilesPerRow, tileIdx / pc.tilesPerRow) * TILE_SIZE;
+    const uint32_t2 localPos = uint32_t2(localIdx % TILE_SIZE, localIdx / TILE_SIZE);
+    const uint32_t2 pixelPos = tileBase + localPos;
+
+    vk::RawBufferStore<uint32_t>(pc.deviceBufferAddress + gIdx * 4u, uint32_t(packUnorm4x8(dstImage[pixelPos])));
+}
+
+[numthreads(128, 1, 1)]
+[shader("compute")]
+void MortonStore(uint32_t3 ID : SV_DispatchThreadID)
```
Erfan-Ahmadi: - make the workgroup 2d 16x16 or a 512 1D workgroup
- but it needs to handle 16x16 region of the tile
- your read pos stays the same with flattened global idx (we need to make sure you're reading contigously)
- `morton::code<false, 7, 2> mc;` you now only need 4 bits for 16x16 so it becomes 4,2 I think
- use the bitshift and & for division and modulo like my prev commit.
   - it's very likely the compiler already does this optimization for you since TILE_SIZE is a macro, but it's good practice, in case it changed later to a push constant or something not known at compile time
- make sure this change is reflected on your dispatch;
  - since each tile is 128x128, it'll take 64 workgroups of size 512 to handle copy of a single tile for you
  - use morton code locally within this 16x16 to figure out the write location + add offset. 
  - doing morton on 16x16 tiles with added offset is no different than doing morton globally on a 128x128 tile (see image below)

Mortong Example for 16x16 group:  
thread 0: reads(0,0) at location 0 writes to pixelPos(0,0)
thread 1: reads(1,0) at location `1*ByteSize` writes to pixelPos(1,0)
thread 2: reads(0,1) at location `2*ByteSize` writes to pixelPos(0,1)
thread 3: reads(1,1) at location `3*ByteSize` writes to pixelPos(1,1)
thread 4: reads(2,0) at location `4*ByteSize` writes to pixelPos(2,0)
...
make sure this is what happes, reads are contigous, writes are morton ordered
might be actually easier to achieve this with single 1D 512 workgroup, not sure

Image

Erfan-Ahmadi: btw we're not going to go with morton benchmarking any further, but let's just fix it up and make it work as we first intended.

73_ImageUploadBenchmark/main.cpp
```diff
@@ -0,0 +1,1572 @@
...
+		m_logger->log("    Memcpy time:       %.3f s (%.1f%%)", ILogger::ELL_PERFORMANCE, totalMemcpyTime, 100.0 * totalMemcpyTime / elapsedSeconds);
+		m_logger->log("    Image create time: %.3f s (%.1f%%)", ILogger::ELL_PERFORMANCE, totalImageCreateTime, 100.0 * totalImageCreateTime / elapsedSeconds);
+		m_logger->log("    Record time:       %.3f s (%.1f%%)", ILogger::ELL_PERFORMANCE, totalRecordTime, 100.0 * totalRecordTime / elapsedSeconds);
+		m_logger->log("    Submit time:       %.3f s (%.1f%%)", ILogger::ELL_PERFORMANCE, totalSubmitTime, 100.0 * totalSubmitTime / elapsedSeconds);
+		m_logger->log("    Memcpy speed:      %.2f GB/s", ILogger::ELL_PERFORMANCE, (totalBytes / (1024.0 * 1024.0 * 1024.0)) / totalMemcpyTime);
+
+		return throughputGBps;
+	}
+
+	BenchResult runBenchmarkCompute(
```
Erfan-Ahmadi, line 760: you probably need to pass in workgroup size here as well

73_ImageUploadBenchmark/main.cpp
```diff
@@ -0,0 +1,1368 @@
...
+		uint32_t tileSize,
+		uint32_t tileSizeBytes,
+		uint32_t tilesPerFrame,
+		uint32_t framesInFlight,
+		uint32_t totalFrames,
+		IQueue* queue)
+	{
+		// Disabled after testing: this path needs CPU writes into host-visible
+		// OPTIMAL images, but the memory layout and preinitialized-image lifetime
+		// rules are too implementation-dependent to make this a clean benchmark.
```
Erfan-Ahmadi: add: "The devices we tested on didn't allow creating OPTIMAL images over host visible memory"

73_ImageUploadBenchmark/main.cpp
```diff
@@ -0,0 +1,1368 @@
...
+		IDeviceMemoryAllocator::SAllocation& stagingAlloc,
+		void* mappedPtr,
+		IGPUImage* destinationImage,
+		IGPUComputePipeline* pipeline,
+		IGPUPipelineLayout* pipelineLayout,
+		IGPUDescriptorSet* ds,
+		uint32_t tileSize,
+		uint32_t tileSizeBytes,
+		uint32_t workgroupSizeX,
+		uint32_t workgroupSizeY,
```
Erfan-Ahmadi: you  could use uint32_t2
it's in nbl::hlsl namespace. makes your life easier with 2D/3D params

73_ImageUploadBenchmark/app_resources/tile_upload.comp.hlsl
```diff
@@ -0,0 +1,99 @@
...
+    const uint32_t2 pixelPos = pc.dstOffset + (dstTile << TILE_SIZE_LOG2) + localPos;
+
+    const uint32_t packed = vk::RawBufferLoad<uint32_t>(pc.deviceBufferAddress + srcPixelIdx * 4u);
+
+    dstImage[pixelPos] = unpackUnorm4x8(int32_t(packed));
+}
+
+[numthreads(128, 4, 1)]
+[shader("compute")]
+void SnakeLoad(uint32_t3 ID : SV_DispatchThreadID)
```
Erfan-Ahmadi: remove the unused load function. I don't think they are used anymore, right?

73_ImageUploadBenchmark/app_resources/tile_upload.comp.hlsl
```diff
@@ -0,0 +1,99 @@
...
+    const uint32_t packedDstTile = vk::RawBufferLoad<uint32_t>(pc.dstTileLocationsAddress + tileIdx * 4u);
+    const uint32_t2 dstTile = uint32_t2(packedDstTile & 0xffffu, packedDstTile >> 16u);
+    const uint32_t2 pixelPos = pc.dstOffset + (dstTile << TILE_SIZE_LOG2) + localPos;
+
+    vk::RawBufferStore<uint32_t>(pc.deviceBufferAddress + dstPixelIdx * 4u, uint32_t(packUnorm4x8(dstImage[pixelPos])));
+}
+
+[numthreads(16, 16, 1)]
+[shader("compute")]
+void MortonStore(uint32_t3 ID : SV_GroupThreadID, uint32_t3 GroupID : SV_GroupID)
```
Erfan-Ahmadi: confusing semantics, you're using ID for GroupThreadID here, but above in SnakeStore you're using the same name for global disaptch ID

PR #264, opened 2026-03-31
Fix EX31 warmup, rectangle area, and RWMC runtime path

AnastaZIuk:
## Summary

- make EX31 warmup and prepared-shader cache preparation line up with the branch's effective persistent render variants so edited HLSL no longer reuses stale prepared SPIR-V
- make rectangle `Area` selection resolve to a real packaged variant inside the existing variant graph instead of a UI choice that can drift from effective selection
- realign EX31 RWMC startup, resolve configuration, and runtime selection with the persistent-workgroup model used on `path_tracer_beauty`
- reset EX31 RWMC defaults to a stable comparison baseline and expose a one-click UI reset so RWMC and non-RWMC brightness checks start from the same parameter state

## Root cause

The current EX31 branch has two separate regressions.

First, the startup and warmup path stopped following the branch's effective packaged variants. The initial runtime-side startup selection came in [`25d3f009`](https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/commit/25d3f009), and [`3b9b2f32`](https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/commit/3b9b2f32) widened that path by preloading the startup shader and resolve shader before the first submit. After the later variant reshapes in [`13f09209`](https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/commit/13f09209), [`520e26fb`](https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/commit/520e26fb), and [`7e0e7f98`](https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/commit/7e0e7f98), EX31 still keyed startup and warmup from the UI-side `usePersistentWorkGroups` bit instead of the branch's effective packaged-variant contract. The same [`7e0e7f98`](https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/commit/7e0e7f98) change also introduced `render_variant_info.hpp` with rectangle hard-wired to `SOLID_ANGLE`, so the UI could expose `Area` while effective selection never reached a real rectangle-area packaged variant.

Second, RWMC drifted in a later same-day refactor sequence. [`2e90b9f0`](https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/commit/2e90b9f0) replaced the old scramble-sequence path with `CCachedOwenScrambledSequence.hpp` and `KeyedQuantizedSequence.hlsl`, and that commit already notes that the new HLSL decoder was broken. [`02717101`](https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/commit/02717101) moved EX31 onto that decoder and resynced the main push constants, but the resolve-side `rwmcParams.sampleCount` no longer followed the active `spp`. [`5d3b3b27`](https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/commit/5d3b3b27) then adapted EX31 after persistent-workgroup and RWMC breakage showed up, but runtime selection, resolve configuration, and the compiled entrypoint surface still did not line up cleanly. This PR makes those paths agree again around the persistent packaged variants the branch actually uses.

The RWMC-parameter brightness drift is older than this branch. [`64ba310b`](https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/commit/64ba310b) introduced the long-lived EX31 RWMC default pack with `kappa = 5.0`, and the same defaults remained on `master`. That was not the cause of the branch runtime regressions above, but it made RWMC versus non-RWMC comparisons depend too much on leftover UI state. This PR also resets EX31 to a stable comparison baseline and adds a direct `Reset RWMC` control so those comparisons are reproducible.


AnastaZIuk:
CI failing due to https://github.com/microsoft/DirectXShaderCompiler/pull/8324, fix about to enter master and we will pull it here

31_HLSLPathTracer/main.cpp
```diff
@@ -1220,17 +1221,9 @@ class HLSLComputePathtracer final : public SimpleWindowedApplication, public Bui
...
-					const auto matT = hlsl::float32_t4x3(m_lightModelMatrix);
-					pc.lightX = matT[0];
-					pc.lightY = matT[1];
-					// Z had a length and a direction, can point colinear or opposite cross product
-					const auto recon = hlsl::cross(matT[0], matT[1]);
-					pc.lightZscale = hlsl::sign(hlsl::dot(recon,matT[2]))*hlsl::length(matT[2])/hlsl::length(recon);
-					pc.lightPos = matT[3];
-					assert(pc.lightMatrix()==hlsl::transpose(matT));
-				}
+				pc.generalPurposeLightMatrix = hlsl::float32_t3x4(transpose(m_lightModelMatrix));
```
devshgraphicsprogramming, line 1224: I did this for 2 reasons.

1. When you `useRWMC` the push constants grow past 128 bytes!
2. the light may not have a skew, so storing more than 6 coordinates makes no sense (I store 7 because reconstruction of the matrix from a quaternion per-invocation is painful)

31_HLSLPathTracer/app_resources/hlsl/render_common.hlsl
```diff
@@ -12,28 +12,14 @@ NBL_CONSTEXPR uint32_t MaxSamplesLog2 = MAX_SAMPLES_LOG2;
...
-    }
-
-    uint64_t pSampleSequence;
 	float32_t4x4 invMVP;
-    float32_t3 lightX;
-    float32_t3 lightY;
-    float32_t lightZscale;
-    float32_t3 lightPos;
+	float32_t3x4 generalPurposeLightMatrix;
+    uint64_t pSampleSequence;
```
devshgraphicsprogramming, line 17: I neeed to undo this when merging (not a big deal)

PR #275, opened 2026-04-25
Sobol sampler benchmarks

karimsayedre:


75_SobolBenchmark/app_resources/shaders/sobol_bench.comp.hlsl
```diff
@@ -0,0 +1,102 @@
...
+   const uint16_t rank = _static_cast<uint16_t>(bitReverse<uint32_t>(invID) >> 16);
+
+   uint16_t seed = _static_cast<uint16_t>(0xdeadu);
+   // simulate rays running paths to exhaustion in a working set
+   [[loop]]
+   for (uint16_t s = _static_cast<uint16_t>(0u); s < Samples; s++)
+   {
+      [[loop]]
+      for (uint16_t d = _static_cast<uint16_t>(0u); d < Depth; d++)
+      {
```
devshgraphicsprogramming: might want to add a mode where the depths are slightly staggered per invocation.

Explanation, SER and other things like wavefront path tracing may pack ray payloads at different depths into invocations of the same subgroup, the difference won't be wild, but it will be enough for you to have to account for having to load different matrices per invocation (which tensor cores really don't like and you need to strip with subgroup ops)

You need to add push constant dependent probability of random (depending on the `triple[0].z` value) `break` out of the depth loop to simulate paths terminating early either due to miss, absorption or russian roulette and not getting refilled. IMHO best would be a linear PDF, e.g. probability at depth=0 and probability at depth=max

devshgraphicsprogramming: @karimsayedre done?

karimsayedre: Oh yeah sorry I did it and forgot to comment

PR #282, opened 2026-05-18
Cad markers optimization

Przemog1:


62_CAD/shaders/globals.hlsl
```diff
@@ -550,11 +583,7 @@ NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t ImagesBindingArraySize = 128;
 NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t MainObjectIdxBits = 24u; // It will be packed next to alpha in a texture
 NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t AlphaBits = 32u - MainObjectIdxBits;
 NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t MaxIndexableMainObjects = (1u << MainObjectIdxBits) - 1u;
```
Erfan-Ahmadi: Can you make the Alpha+MainObject packing a bitfield struct too? and use it in the fragment shader interlock stuff?

62_CAD/DrawResourcesFiller.h
```diff
@@ -868,19 +868,19 @@ struct DrawResourcesFiller
 		
 		// Invalidate all the clip projection addresses because activeProjections buffer got reset
 		for (auto& addr : activeClipRectIndices)
-			addr = InvalidCustomClipRectIndex;
+			addr = MainObject::getInvalidCustomClipRectIndex();
```
Erfan-Ahmadi: Yes, custom clip tect or linestyle is only indexed by main object currently. but it may not stay that way, that's why I suggested this `IndexLimits` previously on discord.
you correctly implemented it but it needs to be outside mainobject


```c++
// make sure to add some good comment here for future maintainability and how we're supposed to expose IndexLimits (by minimum representation).
constexpr inline MinimumLineStyleBits = MainObject::LineStyleBitfield; // currently only referenced by main object so we'll use  
using LineStyleIndexLimits = IndexLimits<uint32_t, LineStyleIdxOrDtmSettingsIdxField::BitCount>
```

Erfan-Ahmadi: I prefer your code this way btw, by asking the main object guy which directly indexes the linestyle. if I knew the code would stay this way. I'd want the limits to be queried from MainObject itself.
but I have a feeling this will change with future compute optimizations and gpu driven rendering

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -2076,8 +2076,9 @@ uint32_t DrawResourcesFiller::addLineStyle_Internal(const LineStyleInfo& lineSty
 {
 	const size_t remainingResourcesSize = calculateRemainingResourcesSize();
 	const bool enoughMem = remainingResourcesSize >= sizeof(LineStyle); // enough remaining memory for 1 more linestyle?
-	if (!enoughMem)
-		return InvalidStyleIdx;
+	const bool indexLimitExceeded = resourcesCollection.lineStyles.vector.size() > MainObject::getMaxIndexableLineStyles();
```
Erfan-Ahmadi: Very Good I didn't expect you to remember adding this part. I was ready to comment that you missed it :D
Good catch.

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -2098,9 +2099,10 @@ uint32_t DrawResourcesFiller::addDTMSettings_Internal(const DTMSettingsInfo& dtm
 	const size_t noOfLineStylesRequired = ((dtmSettingsInfo.mode & E_DTM_MODE::OUTLINE) ? 1u : 0u) + dtmSettingsInfo.contourSettingsCount;
 	const size_t maxMemRequired = sizeof(DTMSettings) + noOfLineStylesRequired * sizeof(LineStyle);
 	const bool enoughMem = remainingResourcesSize >= maxMemRequired; // enough remaining memory for 1 more dtm settings with 2 referenced line styles?
+	const bool indexLimitExceeded = resourcesCollection.dtmSettings.vector.size() > MainObject::getMaxIndexableDtmSettings();
```
Erfan-Ahmadi: Looking at this confusing comparison operation, I realized it was dumb to have "MaxIndexable". 
because us as developers are usually used to comparing sizes. and here it's a bit more confusing because we're guarding against adding one more.

I changed my mind, can we have `MaxCount=Invalid` instead of `MaxIndexable=Invalid-1` and make these comparisons `>=` instead?

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -2152,15 +2154,15 @@ uint32_t DrawResourcesFiller::addDTMSettings_Internal(const DTMSettingsInfo& dtm
 	return resourcesCollection.dtmSettings.addAndGetOffset(dtmSettings); // this will implicitly increase total resource consumption and reduce remaining size --> no need for mem size trackers
 }
 
-float64_t3x3 DrawResourcesFiller::getFixedGeometryFinalTransformationMatrix(const float64_t3x3& transformation, TransformationType transformationType) const
+float64_t2x3 DrawResourcesFiller::getFixedGeometryFinalTransformationMatrix(const float64_t2x3& transformation, TransformationType transformationType) const
 {
 	if (!activeProjections.empty())
 	{
-		float64_t3x3 newTransformation = nbl::hlsl::mul(activeProjections.back(), transformation);
+		float64_t3x3 newTransformation = nbl::hlsl::mul(hlsl::math::linalg::promote_affine<3,3,2,3>(activeProjections.back()), hlsl::math::linalg::promote_affine<3, 3, 2, 3>(transformation));
```
Erfan-Ahmadi, line 2161: Interesting, what do these`hlsl::math::linalg::promote_affine` actually do?
could we actually have the type be "affineTransformation" and avoid using general `float64_t2x3`? this would help with readability and compile time type checks, making sure the inputs are affine transformations,

Just add a TODO for this somewhere relevant to tackle later.

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -2263,7 +2265,7 @@ uint32_t DrawResourcesFiller::acquireActiveMainObjectIndex_SubmitIfNeeded(SInten
 	if (needsCustomClipRect) memRequired += sizeof(WorldClipRect);
 
 	const bool enoughMem = remainingResourcesSize >= memRequired; // enough remaining memory for 1 more dtm settings with 2 referenced line styles?
-	const bool needToOverflowSubmit = (!enoughMem) || (resourcesCollection.mainObjects.vector.size() >= MaxIndexableMainObjects);
+	const bool needToOverflowSubmit = (!enoughMem) || (resourcesCollection.mainObjects.vector.size() > MaxIndexableMainObjects);
```
Erfan-Ahmadi: good catch! 
See? I was confused when writing this  https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/282/changes#r3378075857

Do you agree?

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -2275,26 +2277,41 @@ uint32_t DrawResourcesFiller::acquireActiveMainObjectIndex_SubmitIfNeeded(SInten
...
+	{
+		mainObject.setDtmSettingsIndex(acquireActiveDTMSettingsIndex_SubmitIfNeeded(intendedNextSubmit));
+		mainObject.setDtmSettingsFlag(true);
+	}
+	else
+	{
+		mainObject.setLineStyleIndex(MainObject::getInvalidLineStyleIndex()); // line style and dtm settings indices share the same memory, so no need to invalidate dtm settings here
+		mainObject.setDtmSettingsFlag(false);
+	}
+	mainObject.customTransformationIndex = (needsCustomProjection) ? acquireActiveCustomProjectionIndex_SubmitIfNeeded(intendedNextSubmit) : MainObject::getInvalidCustomTransformationIndex();
```
Erfan-Ahmadi: use a set method, even if it's not packed
consistency is important

62_CAD/shaders/globals.hlsl
```diff
@@ -606,7 +669,19 @@ DTMSettings loadDTMSettings(const uint32_t index)
 }
 pfloat64_t3x3 loadCustomProjection(const uint32_t index)
 {
-    return vk::RawBufferLoad<pfloat64_t3x3>(globals.pointers.customProjections + index * sizeof(pfloat64_t3x3), 8u);
+    pfloat64_t2x3 matrix2x3 = vk::RawBufferLoad<pfloat64_t2x3>(globals.pointers.customProjections + index * sizeof(pfloat64_t2x3), 8u);
```
Erfan-Ahmadi: Ok I know we only use this to mul with the default 3x3 matrix and you need to add that row for the `mul` to work. (In 3D it would be the same problem, affine would be 3x4 and projections would be 4x4)

but since the original customProjection is affine 2x3, I want the load to return it directly and not alter it.
handle the mul in a specific function that does this before the mul. possibly in vertex_shader.hlsl

we might need to (in the future) look into adding some functionalities for affine transformation mul.
can't we use `hlsl::math::linalg::promote_affine` or something similar here?

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -2275,26 +2277,41 @@ uint32_t DrawResourcesFiller::acquireActiveMainObjectIndex_SubmitIfNeeded(SInten
 	MainObject mainObject = {};
 	// These 3 calls below shouldn't need to Submit because we made sure there is enough memory for all of them.
 	// if something here triggers a auto-submit it's a possible bug with calculating `memRequired` above, TODO: assert that somehow?
-	mainObject.styleIdx = (needsLineStyle) ? acquireActiveLineStyleIndex_SubmitIfNeeded(intendedNextSubmit) : InvalidStyleIdx;
-	mainObject.dtmSettingsIdx = (needsDTMSettings) ? acquireActiveDTMSettingsIndex_SubmitIfNeeded(intendedNextSubmit) : InvalidDTMSettingsIdx;
-	mainObject.customProjectionIndex = (needsCustomProjection) ? acquireActiveCustomProjectionIndex_SubmitIfNeeded(intendedNextSubmit) : InvalidCustomProjectionIndex;
-	mainObject.customClipRectIndex = (needsCustomClipRect) ? acquireActiveCustomClipRectIndex_SubmitIfNeeded(intendedNextSubmit) : InvalidCustomClipRectIndex;
-	mainObject.transformationType = (uint32_t)activeMainObjectTransformationType;
+
+	assert((needsLineStyle == !needsDTMSettings) || (needsLineStyle == false && needsDTMSettings == false));
```
Erfan-Ahmadi: `assert(!needsLineStyle || !needsDTMSettings); // both cannot be true at the same time, at least one must be false`

62_CAD/shaders/globals.hlsl
```diff
@@ -144,24 +147,88 @@ enum TransformationType
...
+    static uint32_t getMaxIndexableCustomTransformations() { return nbl::hlsl::numeric_limits<uint32_t>::max - 1u; }
 };
 
 struct DrawObject
 {
-    uint32_t type_subsectionIdx; // packed two uint16 into uint32
-    uint32_t mainObjIndex;
-    uint64_t geometryAddress;
+    uint32_t packedData;
+    uint32_t geometryAddress; // geometry has 8 byte alignment, max 34GB of data to reference
```
Erfan-Ahmadi: remember when they say 32-bit systems can only have 4GB ram max?
that's because 32-bit can only store 4GB. But why did I say 34GB here?

I know geometry address is 8 byte aligned. so instead of holding `0, 8, 16, 24, ... ,` offsets I decide to div by 8 when storing and multiply by 8 when loading. that's what allows 34GB of data to reference which is more than  enough.

So here what I need you to do:
1. Div by 8 on a `bool setGeometryAddress(uint64_t)` function (will return false if it's not 8-byte aligned)
2. Mul by 8 on `uint64_t getGeometryAddress` function
3. We already guarantee this but assert in DRF.cpp that the address is 8-byte aligned by checking return value of `setGeometryAddress` is `true`. (+ use m_logger.log, to log error, would be useful when bug hunting on production builds)

(tbh I don't see the geometry section of the buffer exceeding 1GB anytime soon on n4ce, but still...)

62_CAD/shaders/globals.hlsl
```diff
@@ -547,15 +614,11 @@ inline bool operator==(const DTMSettings& lhs, const DTMSettings& rhs)
...
 NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t ImagesBindingArraySize = 128;
-NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t MainObjectIdxBits = 24u; // It will be packed next to alpha in a texture
-NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t AlphaBits = 32u - MainObjectIdxBits;
-NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t MaxIndexableMainObjects = (1u << MainObjectIdxBits) - 1u;
-NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t InvalidStyleIdx = nbl::hlsl::numeric_limits<uint32_t>::max;
-NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t InvalidDTMSettingsIdx = nbl::hlsl::numeric_limits<uint32_t>::max;
-NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t InvalidMainObjectIdx = MaxIndexableMainObjects;
-NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t InvalidCustomProjectionIndex = nbl::hlsl::numeric_limits<uint32_t>::max;
-NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t InvalidCustomClipRectIndex = nbl::hlsl::numeric_limits<uint32_t>::max;
+NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t MaxIndexableMainObjects = IndexLimits<uint32_t, PseudoStencil::MainObjectIdxField::BitCount>::MaxIndexable;
```
Erfan-Ahmadi: expose both as single
`MainObjectIndexLimits`
also take into account how DrawObject indexes mainObject (you should use the minimum bitcount of the two)
I know pseudoStencil is the minimum, but it somehow needs to be documented by first creating a minimum constexpr value and setting it to pseudostencil + additional comments

62_CAD/shaders/main_pipeline/fragment_shader_resolve_alphas.hlsl
```diff
@@ -19,17 +19,17 @@ float32_t4 calculateFinalColor<true>(const uint2 fragCoord)
     nbl::hlsl::spirv::beginInvocationInterlockEXT();
     
     bool resolve = false;
-    uint32_t toResolveStyleIdx = InvalidStyleIdx;
-    const uint32_t packedData = pseudoStencil[fragCoord];
-    const uint32_t storedQuantizedAlpha = nbl::hlsl::glsl::bitfieldExtract<uint32_t>(packedData,0,AlphaBits);
-    const uint32_t storedMainObjectIdx = nbl::hlsl::glsl::bitfieldExtract<uint32_t>(packedData,AlphaBits,MainObjectIdxBits);
+    uint32_t toResolveStyleIdx = MainObject::getInvalidLineStyleIndex();
```
Erfan-Ahmadi: Code make sense and should work. but let's test if it works
To make sure this works as expected we need to see if we can blend multiple polylines on top of another with non trivial alpha.
run one of the test cases that has many overlapping lines/curves and set their alpha to something that causes blending between different polylines.
make sure the test also includes some static images
check if the transparency algorithm works as expected
resolve this comment if the test was successfull

62_CAD/shaders/main_pipeline/fragment_shader_resolve_alphas.hlsl
```diff
@@ -38,9 +38,17 @@ float32_t4 calculateFinalColor<true>(const uint2 fragCoord)
         // sampling from colorStorage needs to happen in critical section because another fragment may also want to store into it at the same time + need to happen before store
         if (resolve)
         {
-            toResolveStyleIdx = loadMainObject(storedMainObjectIdx).styleIdx;
-            if (toResolveStyleIdx == InvalidStyleIdx) // if style idx to resolve is invalid, then it means we should resolve from color
+            MainObject mainObject = loadMainObject(storedMainObjectIdx);
+            if(!mainObject.isUsingDtmSettings())
```
Erfan-Ahmadi: I think this is a bug when drawing images.

we potentially resovle with color even when not using dtms. for example when drawing images (invalid dtm and line style)

checking whether we have "DTM Rendering" or not is not the main point.
you should check whether it uses `lineStyleColor` to resolve the color.

62_CAD/shaders/globals.hlsl
```diff
@@ -144,24 +147,88 @@ enum TransformationType
...
-    uint32_t customProjectionIndex;
-    uint32_t customClipRectIndex;
-    uint32_t transformationType; // todo pack later, it's just 2 possible values atm
+    uint32_t packedData;
+    uint32_t customTransformationIndex; // needs at least 24 bits
+
+    using LineStyleIdxOrDtmSettingsIdxField = utils::BitField<uint32_t, 0, 16>; // 65,536 distinct lineStyles or dtmSettings is more than enough for an n4ce frame, but make sure auto submit in drawresources filler plays nice and doesn't exceed this value
+    using CustomClipRectIndexField = utils::BitField<uint32_t, 16, 14>; // these are associated with the number of clipping rects or dwgs one could have in a frame. from experience they'll always be less than 10, 32,768 is more than enough
+    using TransformationTypeField = utils::BitField<uint32_t, 30, 1>; // todo pack later, it's just 2 possible values atm
+    using UseDtmSettingsField = utils::BitField<uint32_t, 31, 1>; // this bit indicates if MainObject uses DTM settings
```
Erfan-Ahmadi: instead of "UseDTM" write "ResolveColorFromLineStyle"
It's more descriptive and prevents bugs like this:
https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/282/changes#r3378247187

62_CAD/shaders/main_pipeline/fragment_shader_resolve_alphas.hlsl
```diff
@@ -38,9 +38,17 @@ float32_t4 calculateFinalColor<true>(const uint2 fragCoord)
...
+            {
+                toResolveStyleIdx = mainObject.getLineStyleIndex();
+                if (toResolveStyleIdx == MainObject::getInvalidLineStyleIndex()) // if style idx to resolve is invalid, then it means we should resolve from color
+                    color = float32_t4(unpackR11G11B10_UNORM(colorStorage[fragCoord]), 1.0f);
```
Erfan-Ahmadi: Ok it's not a bug but it's a redundant check and doing the colorStorage read twice in different parts of the code

1. when adding an image main object, needsLineStyle and needsDTMSettings are both false for an image. 
2. then you go inside the first loop, realize the lineStyle index is invalid and read from color storage.

BUT

but why not just do `if(!mainObject.colorFromLineStyle()) color = ...` it's less reads and a single check and less branches.
and you only set that bit to true when you have valid linestyle. 

with this you'll avoid doing `if (toResolveStyleIdx == MainObject::getInvalidLineStyleIndex())` because DRF will guarantee valid style index if you have "colorFromLineStyle"

62_CAD/shaders/main_pipeline/fragment_shader.hlsl
```diff
@@ -101,30 +107,39 @@ float32_t4 calculateFinalColor<true>(const uint2 fragCoord, const float localAlp
...
     // sampling from colorStorage needs to happen in critical section because another fragment may also want to store into it at the same time + need to happen before store
     if (resolve)
     {
-        toResolveStyleIdx = loadMainObject(storedMainObjectIdx).styleIdx;
-        if (toResolveStyleIdx == InvalidStyleIdx) // if style idx to resolve is invalid, then it means we should resolve from color
+        if(!mainObject.isUsingDtmSettings())
+        {
+            toResolveStyleIdx = mainObject.getLineStyleIndex();
+            if (toResolveStyleIdx == MainObject::getInvalidLineStyleIndex()) // if style idx to resolve is invalid, then it means we should resolve from color
+                color = float32_t4(unpackR11G11B10_UNORM(colorStorage[fragCoord]), 1.0f);
```
Erfan-Ahmadi: same comment: https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/282/changes#r3378313495

62_CAD/DrawResourcesFiller.cpp
```diff
@@ -2275,26 +2277,41 @@ uint32_t DrawResourcesFiller::acquireActiveMainObjectIndex_SubmitIfNeeded(SInten
...
+	if (needsLineStyle)
+	{
+		mainObject.setLineStyleIndex(acquireActiveLineStyleIndex_SubmitIfNeeded(intendedNextSubmit));
+		mainObject.setDtmSettingsFlag(false);
```
Erfan-Ahmadi: get rid of setDTMSettingsFlag and
use something like "ResolveColorFromLineStyle" and set only when `needsLineStyle` is true
regarding this:
https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/282/changes#r3378256545

62_CAD/shaders/main_pipeline/fragment_shader_resolve_alphas.hlsl
```diff
@@ -62,7 +70,7 @@ float32_t4 calculateFinalColor<true>(const uint2 fragCoord)
 
     // draw with previous geometry's style's color or stored in texture buffer :kek:
     // we don't need to load the style's color in critical section because we've already retrieved the style index from the stored main obj
-    if (toResolveStyleIdx != InvalidStyleIdx) // if toResolveStyleIdx is valid then that means our resolved color should come from line style
+    if (toResolveStyleIdx != MainObject::getInvalidLineStyleIndex()) // if toResolveStyleIdx is valid then that means our resolved color should come from line style
```
Erfan-Ahmadi: addditional note: this branch will go through if your main object to resolve had "resolveFromLineStyle" set to true

62_CAD/DrawResourcesFiller.cpp
Przemog1: `promote_affine` adds rows/columns to an input matrix, it adds rows/colums with zeros everywhere and ones in the diagonal

62_CAD/DrawResourcesFiller.cpp
Przemog1: yes, "count" is better

62_CAD/shaders/globals.hlsl
Przemog1: > can't we use hlsl::math::linalg::promote_affine or something similar here?

there is no implementation of `hlsl::math::linalg::promote_affine` for `pfloatNxM` matrices

62_CAD/shaders/globals.hlsl
Przemog1: I don't get it

Erfan-Ahmadi: constexpr uint32_t MainObjectMinBitCount = PseudoStencil::MainObjectIdxField::BitCount; // using the bit count in pseudo stencil instead of DrawObj bitfield, because it's the lowest bitcount reference to mainObject and more limiting
using MainObjectIndexLimits = IndexLimits<uint32_t,  MainObjectMinBitCount>;

62_CAD/shaders/main_pipeline/fragment_shader_resolve_alphas.hlsl
Przemog1: so you want
```cpp
// when this function returns true then it can be assumed that `getLineStyleIndex()` will return a valid index to a line style data
    bool colorFromLineStyle() { return (bool)ResolveColorFromLineStyle::get(packedData); }
```

62_CAD/shaders/main_pipeline/fragment_shader_resolve_alphas.hlsl
Przemog1: not a bug as your later comments explain

62_CAD/shaders/main_pipeline/fragment_shader_resolve_alphas.hlsl
Erfan-Ahmadi: Yea

62_CAD/shaders/main_pipeline/vertex_shader.hlsl
```diff
@@ -230,11 +246,11 @@ PSInput vtxMain(uint vertexID : SV_VertexID)
             if (objType == ObjectType::LINE)
             {
                 pfloat64_t2 points[2u];
-                points[0u] = vk::RawBufferLoad<pfloat64_t2>(globals.pointers.geometryBuffer + drawObj.geometryAddress, 8u);
-                points[1u] = vk::RawBufferLoad<pfloat64_t2>(globals.pointers.geometryBuffer + drawObj.geometryAddress + sizeof(LinePointInfo), 8u);
+                points[0u] = vk::RawBufferLoad<pfloat64_t2>(globals.pointers.geometryBuffer + uint64_t(drawObj.getGeometryAddress()), 8u);
```
Erfan-Ahmadi: `drawObj.getGeometryAddress()` should return uint64_t no need to cast
I'll fix this myself, cuz I need it merged

62_CAD/shaders/globals.hlsl
```diff
@@ -547,15 +618,22 @@ inline bool operator==(const DTMSettings& lhs, const DTMSettings& rhs)
...
-NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t MainObjectIdxBits = 24u; // It will be packed next to alpha in a texture
-NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t AlphaBits = 32u - MainObjectIdxBits;
-NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t MaxIndexableMainObjects = (1u << MainObjectIdxBits) - 1u;
-NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t InvalidStyleIdx = nbl::hlsl::numeric_limits<uint32_t>::max;
-NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t InvalidDTMSettingsIdx = nbl::hlsl::numeric_limits<uint32_t>::max;
-NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t InvalidMainObjectIdx = MaxIndexableMainObjects;
-NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t InvalidCustomProjectionIndex = nbl::hlsl::numeric_limits<uint32_t>::max;
-NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t InvalidCustomClipRectIndex = nbl::hlsl::numeric_limits<uint32_t>::max;
+NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t MaxMainObjectCount = IndexLimits<uint32_t, PseudoStencil::MainObjectIdxField::BitCount>::MaxCount;
+NBL_CONSTEXPR_INLINE_NSPC_SCOPE_VAR uint32_t InvalidMainObjectIdx = IndexLimits<uint32_t, PseudoStencil::MainObjectIdxField::BitCount>::Invalid;
```
Erfan-Ahmadi: Ok to have these but these should be retrieved from `MainObjectIndexLimits` (yes it's the same now)
I'll fix this myself, it's a minor change, I need to merge it

PR #287, opened 2026-07-30
Chained scan unit test

keptsecret:


78_ChainedScanUnitTest/app_resources/shaderCommon.hlsl
```diff
@@ -0,0 +1,19 @@
+#include "app_resources/common.hlsl"
+
+using namespace nbl;
+using namespace hlsl;
+
+[[vk::push_constant]] PushConstantData pc;
+
+struct device_capabilities
+{
+#ifdef TEST_NATIVE
```
Erfan-Ahmadi, line 10: small nitpicking: If you do `{ "TEST_NATIVE", "1" }` to enable this, it should be #if not #ifdef

78_ChainedScanUnitTest/main.cpp
```diff
@@ -0,0 +1,515 @@
...
+					}
+					m_api->endCapture();
+
+					// save cache every now and then	
+					{
+						auto cpu = m_spirv_isa_cache->convertToCPUCache();
+						// Normally we'd beautifully JSON serialize the thing, allow multiple devices & drivers + metadata
+						auto bin = cpu->getEntries().begin()->second.bin;
+						IFile::success_t success;
+						m_spirv_isa_cache_output->write(success, bin->data(), 0ull, bin->size());
```
Erfan-Ahmadi, line 263: I can't seem to get the purpose or benefit of the pipeline cache in this example it looks completely unnecessary.
It gets constantly overwritten
and you could actually create a single pipeline per workgroup conf at runtime and not any sort of pipeline cache.

78_ChainedScanUnitTest/app_resources/shaderCommon.hlsl
devshgraphicsprogramming: indeeed

78_ChainedScanUnitTest/main.cpp
devshgraphicsprogramming: slightly faster startup if shader did not change

78_ChainedScanUnitTest/app_resources/testGlobal.comp.hlsl
```diff
@@ -0,0 +1,205 @@
...
+{
+    template<typename AccessType, typename IndexType>
+    void get(const uint32_t ix, NBL_REF_ARG(AccessType) value)
+    {
+        value = scratch[ix];
+    }
+    template<typename AccessType, typename IndexType>
+    void set(const uint32_t ix, const AccessType value)
+    {
+        scratch[ix] = value;
```
devshgraphicsprogramming, line 28: btw if you want these to be visible to other invocations under Vulkan memory model, you might want to use SPIR-V OpLoad `MakeVisible` and OpStore `MakeAvailable` with `NonPrivatePointer` semantics I think

78_ChainedScanUnitTest/app_resources/testGlobal.comp.hlsl
```diff
@@ -0,0 +1,205 @@
...
+    void get(const IndexType index, NBL_REF_ARG(AccessType) value)
+    {
+        bda::__ptr<T> target = ptr + index;
+        value = target.template deref().load();
+    }
+    template<typename AccessType, typename IndexType>
+    void set(const IndexType index, const AccessType value)
+    {
+        bda::__ptr<T> target = ptr + index;
+        return target.template deref().store(value);
```
devshgraphicsprogramming, line 115: I think our `bda::__ptr` needs a load/store method with make available and visible (+ non private)

Its only atomics where avail and vis + non-privateness is implied

78_ChainedScanUnitTest/main.cpp
```diff
@@ -0,0 +1,515 @@
...
+		// check results
+		bool passed = true;
+		//passed = validateResults<Arithmetic, arithmetic::bit_and<uint32_t>>(itemsPerWG, workgroupCount, subgroupSize, itemsPerInvoc);
+		//passed = validateResults<Arithmetic, arithmetic::bit_xor<uint32_t>>(itemsPerWG, workgroupCount, subgroupSize, itemsPerInvoc) && passed;
+		//passed = validateResults<Arithmetic, arithmetic::bit_or<uint32_t>>(itemsPerWG, workgroupCount, subgroupSize, itemsPerInvoc) && passed;
+		passed = validateResults<Arithmetic, arithmetic::plus<uint32_t>>(itemsPerWG, workgroupCount, subgroupSize, itemsPerInvoc) && passed;
+		//passed = validateResults<Arithmetic, arithmetic::multiplies<uint32_t>>(itemsPerWG, workgroupCount, subgroupSize, itemsPerInvoc) && passed;
+		//passed = validateResults<Arithmetic, arithmetic::minimum<uint32_t>>(itemsPerWG, workgroupCount, subgroupSize, itemsPerInvoc) && passed;
+		//passed = validateResults<Arithmetic, arithmetic::maximum<uint32_t>>(itemsPerWG, workgroupCount, subgroupSize, itemsPerInvoc) && passed;
+
```
devshgraphicsprogramming, line 448: why not check all ops ?

78_ChainedScanUnitTest/main.cpp
Erfan-Ahmadi: ok I did not know how pipeline caches work. I though they are overwritten when passed to each create pipeline.
anyways, feels like 95% of the time required for shaders remains to be shader preprocessing and hlsl->spirv compilation with different defines each run of the test

devshgraphicsprogramming: we can cache that too

